PR #4, opened 2020-11-12
Tweak component docs page

chavlji:
Component docs page is now flattened, so we have all stories and documentation on one page:
Screenshot 2020-11-12 at 10 05 46


mplesec (approved):

f-jauregui (commented):
I really like how this looks! 😎 
My only concern is should we delete package-lock.json?

PR #6, opened 2020-11-13
Start Zen-styles

chavlji:
Add basic scss files for **Zen style guide** (zen-styles in short) and add command to produce dist files


package.json
```diff
@@ -41,11 +43,23 @@
...
     "jest-cli": "^26.6.3",
-    "lit-html": "^1.3.0"
+    "lit-html": "^1.3.0",
+    "mini-css-extract-plugin": "^0.5.0",
+    "node-sass": "^4.9.0",
+    "postcss-loader": "^3.0.0",
+    "sass-loader": "^7.0.1",
+    "style-loader": "^0.23.1",
+    "webpack": "^4.6.0",
+    "webpack-cli": "^3.0.0"
```
f-jauregui, line 62: Outdated packages, some of them a couple of release versions. 

- https://www.npmjs.com/package/mini-css-extract-plugin (1.3.1)
- https://www.npmjs.com/package/node-sass (5.0.0)
- https://www.npmjs.com/package/postcss-loader (4.0.4)
- https://www.npmjs.com/package/sass-loader (10.0.1)
- https://www.npmjs.com/package/style-loader (2.0.0)
- https://www.npmjs.com/package/webpack (5.4.0)
- https://www.npmjs.com/package/webpack-cli (4.2.0)

package.json
```diff
@@ -41,11 +43,23 @@
     "@storybook/theming": "^6.0.28",
     "@storybook/web-components": "^6.0.28",
     "@types/jest": "^26.0.15",
+    "autoprefixer": "^9.0.0",
```
f-jauregui, line 46: Outdated package
https://www.npmjs.com/package/autoprefixer (10.0.2)

package.json
```diff
@@ -41,11 +43,23 @@
...
     "babel-loader": "^8.1.0",
+    "breakpoint-sass": "^2.7.1",
     "concurrently": "^5.3.0",
+    "copy-webpack-plugin": "^6.3.0",
+    "css-loader": "^2.0.0",
```
f-jauregui, line 51: Outdated package
https://www.npmjs.com/package/css-loader (5.0.1)

package.json
```diff
@@ -41,11 +43,23 @@
...
+    "cssnano": "^4.0.5",
```
f-jauregui, line 52: Outdated package
https://www.npmjs.com/package/cssnano (4.1.10)

webpack.config.js
```diff
@@ -0,0 +1,62 @@
...
+const CopyPlugin = require('copy-webpack-plugin');
+
+module.exports = {
+  output: {
+    // Webpack will create js files even though they are not used
+    filename: '[name].bundle.js',
+    chunkFilename: '[name].[chunkhash].chunk.js',
+    // Where the CSS is saved to
+    path: path.resolve(__dirname, 'dist/zen-style'),
+    publicPath: "/dist/zen-style"
```
f-jauregui, line 14: **./dist/zen-style** in both cases

mplesec (approved):

f-jauregui (changes requested):
Really liked the way you've put this through!
Still there are some outdated packages. Can we update them before merge?
Since we're starting we should go with the latest and most reliable.

mplesec:
@chavlji @f-jauregui regarding the dependencies, please set a specific version, it's safer that way and won't lead to random version bumps.

f-jauregui:
> @chavlji @f-jauregui regarding the dependencies, please set a specific version, it's safer that way and won't lead to random version bumps.

This way to put versions it's not meant to be random. The `^` updates minor releases and patches.
If we want to be safer but not fixed we can go with `~`, we will have bugfixes without having to update (and look for) insecure packages in short periods of time.

chavlji:
I went with ~. I'd rather go with ^ though. It shouldn't break anything since there aren't breaking changes. But ok, we'll update it manually more often then.

f-jauregui (approved):

PR #10, opened 2020-11-18
BAN-21: Publish package config

f-jauregui:
Changes in this PR:

- Updated main and module entry files
- Added React and other required peer dependencies for preview
- Updated start and build scripts
- Added important config in order to be publish-ready
- Updated addons order to prevent [this](https://github.com/storybookjs/storybook/issues/11442)


package.json
```diff
@@ -15,8 +15,8 @@
     "loader/"
   ],
   "scripts": {
-    "start": "concurrently --raw -n storybook,stencil -c \"bgMagenta.bold,bgBlue.bold\" \"npm run stencil:watch\" \"npm run storybook:start\"",
```
chavlji, line 18: I've checked this and it works nice. The only minor issue is that stencil gets built twice on yarn start, but 🤷‍♂️

chavlji (approved):
Great job! Works really nice

package.json
f-jauregui: Is compiled first so storybook can grab it, then starts the watcher so it compiles one more time and for any change you do.

PR #11, opened 2020-11-18
BAN-32: Flex utils

chavlji:
Added flex utility classes.

I've also imported `utilities/index.scss` to Storybook, so we can use all utility classes on Storybook design directly. For now, I'd avoid importing the whole `zen-ui-styles` into the Storybook.

Screenshot 2020-11-18 at 15 10 08


f-jauregui (approved):
LGTM :shipit:

PR #13, opened 2020-11-23
BAN-46: New output target removed style bundle

f-jauregui:
Updates in this PR:
- Updated create-component command (`yarn create-component my-component-name`)
- Added a new output target so we can import components one by one with other applications
- Made @stencil/core a non-dev dependency. Required by the dist-custom-elements-bundle output

Also have a quick update for BAN-49: 
- Removed unnecessary dependencies and npm scripts from the styles bundle


mplesec (approved):
Nice 👍

chavlji (approved):
Awesome!

PR #16, opened 2020-11-26
BAN-66: Add open-sans font to storybook

chavlji:
[jira](https://reciprocitylabs.atlassian.net/browse/BAN-66)


.storybook/webpack.config.js
```diff
@@ -0,0 +1,37 @@
...
+    use: [
+      {
+        loader: 'file-loader',
+        query: {
+          name: '[name].[ext]'
+        }
+      }
+    ],
+    include: path.resolve(__dirname, '../')
+  }); */
```
f-jauregui, line 25: What are these comments for?

package.json
```diff
@@ -63,8 +68,9 @@
   },
   "license": "MIT",
   "dependencies": {
-    "lodash": "~4.17.20",
-    "@stencil/core": "~2.3.0"
+    "@stencil/core": "~2.3.0",
+    "copy-webpack-plugin": "^6.3.2",
```
f-jauregui, line 72: Move it to devDependencies

f-jauregui (changes requested):
Not super happy about compiling CSS again, but it's ok to keep track of the css bundle size and we can remove it later.
Then just these little comments below.

package.json
f-jauregui: > "copy-webpack-plugin": "^6.3.2",

.storybook/webpack.config.js
chavlji: 😅

PR #25, opened 2020-12-01
BAN-53: Basic Input with Label and Support Text

f-jauregui:
[Jira ticket](https://reciprocitylabs.atlassian.net/browse/BAN-53)


src/components/zen-input/zen-input.tsx
```diff
@@ -0,0 +1,39 @@
...
+  /** Makes input required */
+  @Prop() required: boolean = false;
+
+  render() {
+    return (
+      <Host>
+        <div class="input-group">
+          {
+          this.label ?
+            <label>
```
chavlji, line 26: I know this is a wip, but just a note here:
Label and supporting text are not part of this component. We said they will be standalone components and wrapped with a 4th component called ZenFormGroup.

chavlji (changes requested):

src/components/zen-input/zen-input.tsx
```diff
@@ -0,0 +1,31 @@
...
+})
+export class ZenInput {
+  /** Placeholder of the input */
+  @Prop() placeholder: string = null;
+
+  /** Disables input */
+  @Prop() disabled = false;
+
+  /** Makes input required */
+  @Prop() required = false;
```
chavlji, line 16: You're not emitting any events. How will consumers know what's the value?
Also pay attention on how to get that value from FromGroup? Shouldn't be problems since we're using slots I guess...

src/components/zen-input/zen-input.stories.ts
```diff
@@ -0,0 +1,39 @@
+import { html } from 'lit-html';
+import markdown from './readme.md';
+
+const argTypes = {};
+
+export default {
+  title: 'Components/Zen Input',
+  component: 'zen-input',
+  argTypes,
```
chavlji, line 9: Please add all/most props to playground, so we can test if everything works directly from storybook.

src/components/zen-input/zen-input.scss
```diff
@@ -0,0 +1,39 @@
...
+    color: $color-gray-400;
+  }
+
+  &:disabled {
+    opacity: 0.5;
+    pointer-events: none;
+    cursor: default;
+  }
+
+  &:hover {
```
chavlji, line 31: This should probabbly go to our input mixin?
Also :focus, :disabled,...

src/components/zen-input/zen-input.stories.ts
```diff
@@ -0,0 +1,39 @@
...
+  parameters: {
+    notes: { markdown },
+  },
+};
+
+const Template = () => {
+  return html` <zen-input placeholder="Insert text here..." /> `;
```
chavlji, line 16: I'm unable to use keyboard keys 'space', 'left arrow', 'right arrow',... inside input in any story?

src/components/zen-input-support-text/zen-input-support-text.scss
```diff
@@ -0,0 +1,6 @@
+@import "../../zen-styles/variables.scss";
+
+span.supporting-text {
+  font-size: 0.75rem;
```
chavlji, line 4: Please use $text-sm

chavlji (changes requested):

chavlji:
One more thing to pay attention to; I think input should have width: 100% by default.

Otherwise consumers can't control the width. With setting input with to 100%, input will follow the container (container = actual component that app can control = `<zen-input style="width: 500px" />`).

We have to keep that in mind for all the components.

src/components/zen-input/zen-input.stories.ts
f-jauregui: Not sure what you mean with this one

chavlji: I open go to input story, focus input, write 'abc123' - all good. I then (still inside input) press `space` - it writes nothing, but scrolls to end of the page.
In story for *form group* same thing happens. And one more: if I press `arrow left/right`, text carret is supposed to move left/right. But it does nothing.
I can record video, but you won't know which key's I'm pressing...

f-jauregui: This was working fine. Resolved.

src/components/zen-input/test/zen-input.spec.tsx
```diff
@@ -0,0 +1,18 @@
+import { newSpecPage } from '@stencil/core/testing';
+import { ZenInput } from '../zen-input';
+
+describe('zen-input', () => {
```
chavlji, line 4: Tests for all 4 components are covering just the default setup and aren't checking props/variants/events.
Add a separate task, to cover that for all 4 components.

src/components/zen-input-label/zen-input-label.tsx
```diff
@@ -0,0 +1,25 @@
+import { Component, h, Host, Prop } from '@stencil/core';
+
+@Component({
```
chavlji, line 3: There's no documentation for zen-input-label nor zen-input-support-text. Nowhere to see available options. Create a task for this please.
Where do we put docs for these micro components? It's kinda ugly to list them separately without a context because there'll be a ton of them, but I can't see any other option atm.

src/components/zen-input-label/zen-input-label.tsx
```diff
@@ -0,0 +1,25 @@
...
+  tag: 'zen-input-label',
```
chavlji, line 4: zen-form-group-label would be better name, because you control can also be dropdown, checkbox, textarea, radios, ...
If you agree, please create a task for this.

src/components/zen-input-label/zen-input-label.tsx
```diff
@@ -0,0 +1,25 @@
...
+  styleUrl: 'zen-input-label.scss',
+  shadow: true,
+})
+export class ZenInputLabel {
+  /** Text of the label */
+  @Prop() readonly text: string = null;
```
chavlji, line 10: name 'label' would be better

src/components/zen-input/zen-input.tsx
```diff
@@ -0,0 +1,63 @@
...
+  private onInput = (ev: Event) => {
+    const input = ev.target as HTMLInputElement | null;
+    if (input) {
+      this.value = input.value || '';
+    }
+    this.zenInput.emit(ev as KeyboardEvent);
+  };
+
+  private getValue(): string {
+    return typeof this.value === 'number' ? this.value.toString() : (this.value || '').toString();
```
chavlji, line 43: shouldn't we rather check input's attr type?

src/components/zen-input/zen-input.tsx
```diff
@@ -0,0 +1,63 @@
...
+
+  /**
+   * Disables input.
+   */
+  @Prop() readonly disabled = false;
+
+  /**
+   * Makes input required.
+   */
+  @Prop() readonly required = false;
```
chavlji, line 22: prop 'type' is missing

src/components/zen-input/zen-input.tsx
```diff
@@ -0,0 +1,63 @@
...
+    return typeof this.value === 'number' ? this.value.toString() : (this.value || '').toString();
+  }
+
+  render(): HTMLElement {
+    const value = this.getValue();
+
+    return (
+      <Host>
+        <input
+          type="text"
```
chavlji, line 52: we should support other types

src/components/zen-input/zen-input.stories.ts
```diff
@@ -0,0 +1,52 @@
...
+  },
+  required: {
+    type: { name: 'string', required: false },
+    control: {
+      type: 'boolean',
+    },
+  },
+  disabled: {
+    type: { name: 'string', required: false },
+    control: {
```
chavlji, line 28: this isn't necessary. It can figure out that it's a boolean prop. You only need to specify some special ones like color picker.

src/components/zen-input/zen-input.stories.ts
```diff
@@ -0,0 +1,52 @@
+import { html } from 'lit-html';
+import markdown from './readme.md';
+import { action } from '../../../.storybook/helpers/custom-action';
+
+const customEvents = ['zenInput'];
+
+const argTypes = {
+  value: {
+    type: { name: 'string', required: false },
```
chavlji, line 9: this isn't needed... It can figure it out automatically.

src/components/zen-input/zen-input.stories.ts
```diff
@@ -0,0 +1,52 @@
+import { html } from 'lit-html';
+import markdown from './readme.md';
+import { action } from '../../../.storybook/helpers/custom-action';
+
+const customEvents = ['zenInput'];
+
+const argTypes = {
```
chavlji, line 7: You can remove whole argTypes obj, since it only has basic stuff that it can automatically figure out from passed default arguments for component.
Check zen-steps stories.

src/zen-styles/mixins/forms.scss
```diff
@@ -1,17 +1,23 @@
...
   color: $color-gray-700;
   background: $color-white;
   box-sizing: border-box;
   border: 1px solid $color-gray-200;
   border-radius: $input-radius;
+  display: block;
+  font-family: inherit;
+  font-size: $text-sm;
+  padding: .375rem .5rem;
+  margin: .5rem 0;
```
chavlji, line 11: Remove these margins. Put it into zen-form-group. Components shouldn't define it's outside layout.

src/zen-styles/mixins/forms.scss
```diff
@@ -26,4 +32,18 @@
   &:active {
     @include zenInputFocus;
   }
+
+  &::placeholder {
```
chavlji, line 36: Nice!

chavlji (approved):
Can be merged, but please create a ticket for the requested in this case and place their links under the comments (to track it better).

src/components/zen-input/zen-input.tsx
f-jauregui: no, this is the default component value

src/components/zen-input-label/zen-input-label.tsx
f-jauregui: zen-label will be better

src/components/zen-input/test/zen-input.spec.tsx
f-jauregui: Yes, we should do that in all components

src/components/zen-input/zen-input.tsx
f-jauregui: Automatically inferred, since is a boolean there is no other valid option. And actually linter will complain about that.

src/components/zen-input/zen-input.tsx
f-jauregui: As requirement comes in will do

src/components/zen-input/zen-input.stories.ts
f-jauregui: argTypes obj defautValues is the way to do it.
https://storybook.js.org/docs/react/api/argtypes

src/components/zen-input/zen-input.stories.ts
f-jauregui: you still need to specify it as argument, and inside argTypes object is the way to do it

src/components/zen-input/test/zen-input.spec.tsx
f-jauregui: https://reciprocitylabs.atlassian.net/browse/BAN-86

src/components/zen-form-group/zen-form-group.stories.ts
```diff
@@ -0,0 +1,66 @@
...
+  parameters: {
+    notes: { markdown },
+  },
+};
+
+const Template = () => {
+  return html`
+    <zen-form-group>
+      <zen-label label="Username" required></zen-label>
+      <zen-input style="margin: .5rem 0;"></zen-input>
```
chavlji, line 19: This margin should be set from form-group.css, because we want to have standardized spacings in all form groups.
You can use `::slotted()` selector for that.

PR #34, opened 2020-12-07
QUICKFIX: Added pinst and updated package version

f-jauregui:
Due to:

Captura de Pantalla 2020-12-07 a la(s) 11 09 09

Added https://github.com/typicode/pinst to prevent postinstall (husky) from run when package is consumed.


leferreyra (approved):
looks legit!

chavlji (approved):
Nice!

PR #35, opened 2020-12-08
BAN-87: Update zen-steps component with last design

chavlji:
See commits and [Jira](https://reciprocitylabs.atlassian.net/browse/BAN-87)


mihaelpalfi (approved):
Good that this is now an option. NIT: When you set that its not clickable the pointer icon should not be displayed.

chavlji:
@mihaelpalfi thought about that yeah. Not that simple so I'll put it into next pr.

f-jauregui (approved):
Great work, @chavlji! 💪

PR #40, opened 2020-12-09
BAN-58: Add tooltip component

mihaelpalfi:
https://reciprocitylabs.atlassian.net/browse/BAN-58

Test cases and stories to the guidelines that Jan will write will be handled here https://reciprocitylabs.atlassian.net/browse/BAN-115


src/components/zen-tooltip/zen-tooltip.tsx
```diff
@@ -0,0 +1,28 @@
...
+  /** Set error state  */
+  @Prop() readonly error?: boolean = false;
+
+  /** Set text */
+  @Prop() readonly text?: string = '';
+
+  render() {
+    return (
+      <Host>
+        <span class={`${this.error ? 'error' : ''} ${this.arrow ? this.arrow : ''}`}>
```
mplesec, line 22: NIT: Move this to a class method, it'll be easier to test.

src/components/zen-tooltip/zen-tooltip.tsx
```diff
@@ -0,0 +1,28 @@
...
+  shadow: true,
+})
+export class ZenTooltip {
+  /** Arrow position */
+  @Prop() readonly arrow?: ArrowPosition = ArrowPosition.LEFT;
+
+  /** Set error state  */
+  @Prop() readonly error?: boolean = false;
+
+  /** Set text */
```
mplesec, line 16: Add clearer descriptions for properties.

src/components/zen-tooltip/zen-tooltip.tsx
```diff
@@ -0,0 +1,28 @@
...
+import { ArrowPosition } from './types';
+
+@Component({
+  tag: 'zen-tooltip',
+  styleUrl: 'zen-tooltip.scss',
+  shadow: true,
+})
+export class ZenTooltip {
+  /** Arrow position */
+  @Prop() readonly arrow?: ArrowPosition = ArrowPosition.LEFT;
```
mplesec, line 11: The default for `arrow` property should probably be bottom since this is how it's displayed in the stories.

src/components/zen-tooltip/types.tsx
```diff
@@ -0,0 +1,6 @@
+export enum ArrowPosition {
+  TOP = 'top',
+  RIGHT = 'right',
+  LEFT = 'left',
+  BOTTOM = 'bottom',
+}
```
mplesec, line 6: Total NIT, but I'd switch the LEFT and BOTTOM so it's consistent with how in i.e. CSS it expects these in clockwise order.
Screenshot 2020-12-09 at 12 33 31

src/components/zen-tooltip/zen-tooltip.tsx
```diff
@@ -0,0 +1,28 @@
...
+
+  /** Set error state  */
+  @Prop() readonly error?: boolean = false;
```
mplesec, line 14: Might be better to use a `variant` property to indicate different state instead of just an error prop.

mplesec (changes requested):
Looking good, needs some fixes. Also didn't review the styles, perhaps someone else should do that since I don't know exactly the preferences around that.

Regarding the arrow position, I think we should use a position property instead. That way the component itself then handles both where the container is positioned and where the arrow is displayed on it. @chavlji @leferreyra opinions?

Regarding stories, also provide a story for how this works on an element. The example should allow user to hover over an element and see the tooltip display. Also the title of each one should be clearer, and probably doesn't need "Story" in it. 😅
Screenshot 2020-12-09 at 12 36 21

src/components/zen-tooltip/zen-tooltip.tsx
f-jauregui: Or you can do the JSX way:

```jsx
<span class={{ 'error': this.error, this.arrow: this.arrow }}>
```

But I won't use `this.arrow`. Citing @mplesec:

> Regarding the arrow position, I think we should use a position property instead. That way the component itself then handles both where the container is positioned and where the arrow is displayed on it. @chavlji @f-jauregui opinions?

I agree with that, we should provide a `@Prop() position: string;` and depending on that the component itself will handle its position and where the arrow is displayed.

src/components/zen-tooltip/zen-tooltip.scss
```diff
@@ -0,0 +1,77 @@
+@import '../../zen-styles/component-helpers.scss';
+
+:host {
+  display: block;
+}
+
+span {
+  background: $color-gray-800;
+  font-size: 0.75rem;
```
f-jauregui, line 9: Use `$text-sm` variable.

src/components/zen-tooltip/zen-tooltip.scss
```diff
@@ -0,0 +1,77 @@
...
+  color: #fff;
```
f-jauregui, line 10: Nit, but $color-white.

src/components/zen-tooltip/zen-tooltip.tsx
f-jauregui: Also when left and right positions: the arrow (and tooltip) should be centered relative to the item that will trigger the tooltip.

src/components/zen-tooltip/zen-tooltip.scss
```diff
@@ -0,0 +1,77 @@
...
+  padding: 0.9375rem;
```
f-jauregui, line 11: Use 1rem.

src/components/zen-tooltip/zen-tooltip.scss
```diff
@@ -0,0 +1,77 @@
...
+  border-radius: 0.1rem;
```
f-jauregui, line 12: You can consume $button-radius variable, but would be better to have a $border-radius-default since the same it's used across different components (And border-radius can be in pixels, for this variable case: 2px)

src/components/zen-tooltip/zen-tooltip.scss
```diff
@@ -0,0 +1,77 @@
...
+  line-height: 1rem;
+  text-align: center;
+  font-weight: 400;
```
f-jauregui, line 15: Not necessary.

src/components/zen-tooltip/zen-tooltip.scss
```diff
@@ -0,0 +1,77 @@
...
+  position: relative;
+  display: inline-block;
+  font-family: inherit;
+  max-width: 11rem;
```
f-jauregui, line 19: I would put this property into :host pseudo-class. In that way, max-width can be modified from outside targeting the just the component.

f-jauregui (changes requested):
Just some tweaks, otherwise it looks really good to me. Good job, @mihaelpalfi!
Another good thing would be to provide a story when the tooltip is displayed along with other component, like a zen-button, so we can see it in action 😉

src/components/zen-tooltip/zen-tooltip.tsx
mihaelpalfi: Done

src/components/zen-tooltip/zen-tooltip.tsx
mihaelpalfi: Done

src/components/zen-tooltip/types.tsx
mihaelpalfi: Done

src/components/zen-tooltip/zen-tooltip.tsx
mihaelpalfi: Done

src/components/zen-tooltip/zen-tooltip.tsx
mihaelpalfi: Moved class method.

src/components/zen-tooltip/zen-tooltip.scss
mihaelpalfi: done

src/components/zen-tooltip/zen-tooltip.scss
mihaelpalfi: done

src/components/zen-tooltip/zen-tooltip.scss
mihaelpalfi: done

src/components/zen-tooltip/zen-tooltip.scss
mihaelpalfi: done

src/components/zen-tooltip/zen-tooltip.scss
mihaelpalfi: removed

mplesec (approved):

f-jauregui (changes requested):
Please remove `src/components/zen-tooltip/assets/profile.svg` it's adding 1.57MB to the final bundle. We should avoid using these asses that won't be used on production, place a dummy CSS circle (maybe with a letter) instead.

src/components/zen-tooltip/zen-tooltip.stories.ts
```diff
@@ -0,0 +1,155 @@
...
+    </div>
+  `;
+};
+
+export const Default = Template.bind({});
+Default.args = {
+  position: Position.TOP,
+  variant: Variant.DARK,
+  text:
+    'The Manager can edit and create in the context of the PCI-DSS framework. They can also grant access to the framework to additional teammates.',
```
chavlji, line 155: nit: Could this example text be max 50 chars long? It requires too much padding in the canvas.

src/components/zen-tooltip/types.tsx
```diff
@@ -0,0 +1,12 @@
+export enum Position {
```
chavlji, line 1: Codebase is becoming cluttered with local type.ts files. Can you please move content of this file into `components/helpers/types.ts`, because variants, aligns, positions,... should have same possible values across all components.

src/components/zen-tooltip/test/zen-tooltip.spec.tsx
```diff
@@ -0,0 +1,18 @@
+import { newSpecPage } from '@stencil/core/testing';
+import { ZenTooltip } from '../zen-tooltip';
+
+describe('zen-tooltip', () => {
+  it('renders', async () => {
+    const page = await newSpecPage({
+      components: [ZenTooltip],
+      html: `<zen-tooltip></zen-tooltip>`,
+    });
+    expect(page.root).toEqualHtml(`
```
chavlji, line 10: Tests are missing. Please add a ticket.

src/components/zen-tooltip/mixins.scss
```diff
@@ -0,0 +1,5 @@
+@mixin tooltip-variant($color, $background, $border) {
```
chavlji, line 1: This mixin only adds complexity and masks what it really does. Remove it and write values directly in scss.

src/components/zen-tooltip/zen-tooltip.scss
chavlji: I still see `border-radius: 0.1rem;`. It should be `border-radius: $border-radius-default` right?

src/components/zen-tooltip/zen-tooltip.scss
```diff
@@ -0,0 +1,127 @@
...
+  position: relative;
+  pointer-events: none;
+  font-size: $text-sm;
+  color: $color-white;
+  padding: 1rem;
+  border-radius: 0.1rem;
+  line-height: 1rem;
+  text-align: center;
+  font-family: inherit;
+  z-index: 999;
```
chavlji, line 22: Let's create a variable `$z-index-tooltips` and add it to `zen-styles/variables.scss`

src/components/zen-tooltip/zen-tooltip.scss
```diff
@@ -0,0 +1,127 @@
...
+}
+
+span:before {
```
chavlji, line 25: Prefere syntax `&:before` inside previous span selector. It better resembles hierarchy of the rules.

src/components/zen-tooltip/zen-tooltip.scss
```diff
@@ -0,0 +1,127 @@
...
+@import './mixins.scss';
+
+:host {
+  position: absolute;
+  display: none;
+  max-width: 11rem;
+  z-index: 999;
+}
+
+span {
```
chavlji, line 11: Don't target element types. Rather append class and target it.

src/components/zen-tooltip/zen-tooltip.scss
```diff
@@ -0,0 +1,127 @@
...
+  line-height: 1rem;
+  text-align: center;
+  font-family: inherit;
+  z-index: 999;
+}
+
+span:before {
+  content: "";
+  position: absolute;
+  width: 0;
```
chavlji, line 28: Why do you need that?

src/components/zen-tooltip/zen-tooltip.tsx
```diff
@@ -0,0 +1,92 @@
...
+export class ZenTooltip {
+  private element: HTMLElement;
+  /** Set tooltip position */
+  @Prop() readonly position?: Position = Position.TOP;
+
+  /** Set tooltip variant */
+  @Prop() readonly variant?: Variant = Variant.DARK;
+
+  /** Set tooltip text */
+  @Prop() readonly text?: string = '';
```
chavlji, line 18: nit: I'd rename it to `label` since all other components use this name (button, dropdown, steps,...)

src/components/zen-tooltip/zen-tooltip.tsx
```diff
@@ -0,0 +1,92 @@
...
+        y += (bounds.height + myBounds.height) / 2 + this.offset;
+        break;
+    }
+
+    this.element.style.left = `${x}px`;
+    this.element.style.top = `${y}px`;
+  }
+
+  hide(): void {
+    this.element.style.display = 'none';
```
chavlji, line 65: Instead of this add `@state visible` and then inside component:
`<Host class={{ visible: this.visible }}>` and add style `.visible { display: none; }`.
It will be more readable and easier to test.

src/components/zen-tooltip/zen-tooltip.scss
```diff
@@ -0,0 +1,127 @@
...
+
+:host {
+  position: absolute;
+  display: none;
+  max-width: 11rem;
+  z-index: 999;
+}
+
+span {
+  display: block;
```
chavlji, line 12: Don't use span with display: block. Use div instead which has display:block by default.

chavlji (changes requested):
Nice! Needs just a bit of fine-tunning.

chavlji:
> Please remove `src/components/zen-tooltip/assets/profile.svg` it's adding 1.57MB to the final bundle. We should avoid using these asses that won't be used on production, place a dummy CSS circle (maybe with a letter) instead.

@f-jauregui why it adds to components bundle size if it's only used in storybook?

src/components/zen-tooltip/test/zen-tooltip.spec.tsx
mihaelpalfi: Look at the description of pr.

src/components/zen-tooltip/zen-tooltip.scss
mihaelpalfi: Waited for notifications to be merged. It had this variable.

src/components/zen-tooltip/zen-tooltip.stories.ts
mihaelpalfi: Changed.

src/components/zen-tooltip/types.tsx
mihaelpalfi: Done

src/components/zen-tooltip/mixins.scss
mihaelpalfi: Removed

src/components/zen-tooltip/zen-tooltip.scss
mihaelpalfi: Done

src/components/zen-tooltip/zen-tooltip.scss
mihaelpalfi: For the arrow to be positioned correctly.

src/components/zen-tooltip/zen-tooltip.tsx
mihaelpalfi: Renamed

src/components/zen-tooltip/zen-tooltip.scss
mihaelpalfi: Done

src/components/zen-tooltip/zen-tooltip.scss
mihaelpalfi: Done.

src/components/zen-tooltip/zen-tooltip.tsx
mihaelpalfi: Done

src/components/zen-tooltip/zen-tooltip.scss
mihaelpalfi: Changed span to div.

src/components/zen-tooltip/zen-tooltip.tsx
chavlji: Great job!

src/components/zen-tooltip/zen-tooltip.scss
```diff
@@ -0,0 +1,121 @@
...
+  display: none;
+}
+
+:host(.visible) {
+  display: block;
+}
+
+.tooltip {
+  display: block;
+  position: relative;
```
chavlji, line 16: nit: You can then remove `display: block`, because div is block by default.

chavlji (approved):
Great!

chavlji (approved):

mplesec (approved):

src/components/zen-tooltip/zen-tooltip.scss
mihaelpalfi: Will do in the ticket mentioned in the description.

f-jauregui (approved):
Great job, Mihael! 🚢

PR #45, opened 2020-12-10
BAN-88: Add notification component

mihaelpalfi:
https://reciprocitylabs.atlassian.net/browse/BAN-88

Closed previous [PR](https://github.com/reciprocity/zen-ui/pull/37) and recreated it here. Added the review requested changes fixes.


src/components/zen-notification/zen-notification.stories.ts
```diff
@@ -0,0 +1,115 @@
+import { html } from 'lit-html';
+import markdown from './readme.md';
+import { ZenDismissDuration, ZenVariant } from './zen-notification-helper';
+
+const argTypes = {
```
f-jauregui, line 5: Arg types that are not type select can be put into a Default.args object, check how zen-step does that, that will inference automatically data type.
Remove description and type from variant and dismissDuration.

src/zen-styles/colors.scss
```diff
@@ -25,6 +25,7 @@ $color-red-100: #fcdadd;
 $color-green-900: #1b422e;
 $color-green-800: #2a6949;
 $color-green-700: #34835b;
+$color-green-600: #3bbf74;
```
f-jauregui, line 28: That green doesn't exist in color palette of the design system. Maybe you can use green-500 or green-700 to achieve that. Should be same color as border-color.

f-jauregui (approved):
Just those little changes.
It's looking better now! 😎

f-jauregui:
Don't forget to create a story to add complete spec tests. Those tests are not enough.

src/components/zen-notification/zen-notification.stories.ts
mihaelpalfi: Done

src/zen-styles/colors.scss
mihaelpalfi: Done

PR #53, opened 2020-12-15
BAN-91: Zen input with slots

f-jauregui:
[Jira ticket](https://reciprocitylabs.atlassian.net/browse/BAN-91)

### DIFF
--- a/src/components/zen-input/zen-input.scss
+++ b/src/components/zen-input/zen-input.scss
@@ -26,6 +26,14 @@ input.input-control {
 
 :host(.invalid) {
   box-shadow: 0 0 0 2px $color-red-300;
+  border: 1px solid $color-red-700;
+}
+
+:host(.disabled) {
+  opacity: 0.5;
+  background-color: $color-gray-100;
+  pointer-events: none;
+  cursor: default;
 }
 
 input {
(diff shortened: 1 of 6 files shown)


mihaelpalfi (approved):
Great work with the slots!

src/components/zen-input/zen-input.scss
```diff
@@ -2,12 +2,57 @@
...
 }
 
-input.input-control {
-  @include zenInput;
-  @include noselect;
+:host(:hover) {
+  border-color: $color-blue-700;
+}
+
+:host(.has-focus) {
```
chavlji, line 14: I think it's great that you're using host to paint the border. I just don't like that we can't use the old mixin for that. We should take a look into that later, to unify it somehow. Shall we create a ticket for that?

chavlji (changes requested):
Looks good! We'll need public method `toggleFocused` exposed so we can turn it on/off when dropdown is un/focused.

src/components/zen-input/zen-input.scss
f-jauregui: I don't think so, if we move forward with Tailwind these mixins won't be needed. I won't fix something that will be replaced sooner than later.

chavlji (approved):
Great! 🚀

PR #63, opened 2020-12-16
BAN-130: Prevent emit click event on disabled button

f-jauregui:
[Jira ticket](https://reciprocitylabs.atlassian.net/browse/BAN-130)

### DIFF
--- a/src/components/zen-button/zen-button.scss
+++ b/src/components/zen-button/zen-button.scss
@@ -6,6 +6,10 @@
   width: fit-content;
 }
 
+:host(.disabled) {
+  pointer-events: none;
+}
+
 .btn {
   position: relative;
   display: flex;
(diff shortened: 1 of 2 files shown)


mihaelpalfi (approved):

chavlji (approved):
Why is this needed? Isn't disabled on the button itself enough?

f-jauregui:
> Why is this needed? Isn't disabled on the button itself enough?

As the app-side event is added to the :host element, the MouseEvent is being triggering even if we have the pointer-events set to none in the inner native button.

chavlji:
A-ok I see. But I think it's not the right approach.

What if we had padding on host. Clicking on such padding would evoke the action which isn't ok. I know it's a bit unrealistic example on a button, but might become an issue with more complex components.

We should bubble click event from the actual button. If button is disabled, there'll be no event. First idea would be to trigger zenClick event on button click and app would hook to that. Maybe it can even be done more elegant with propper event propagation and bubbling.

PR #65, opened 2020-12-17
quickfix: modified button to use standard DOM events for click and enter key

f-jauregui:


chavlji (approved):
Great! A lot cleaner and much less code!

PR #79, opened 2020-12-28
BAN-142: Bringed tests back running transforms with babel-jest

f-jauregui:
[JIRA ticket](https://reciprocitylabs.atlassian.net/browse/BAN-142)

#### Description
As per JIRA ticket

#### ChangeLOG

#### Useful links

### DIFF
--- a/stencil.config.ts
+++ b/stencil.config.ts
@@ -21,4 +21,10 @@ export const config: Config = {
     },
   ],
   plugins: [sass()],
+  testing: {
+    transform: {
+      '^.+\\.js?$': 'babel-jest',
+    },
+    transformIgnorePatterns: ['node_modules/(?!query-selector-shadow-dom/)'],
+  },
 };
--- a/.husky/pre-push
+++ b/.husky/pre-push
@@ -1,4 +1,4 @@
 #!/bin/sh
 . "$(dirname $0)/_/husky.sh"
 
-yarn lint:check && yarn format:check
+yarn lint:check && yarn format:check && yarn test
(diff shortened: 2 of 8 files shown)


stencil.config.ts
```diff
@@ -21,4 +21,10 @@ export const config: Config = {
     },
   ],
   plugins: [sass()],
+  testing: {
+    transform: {
+      '^.+\\.js?$': 'babel-jest',
+    },
+    transformIgnorePatterns: ['node_modules/(?!query-selector-shadow-dom/)'],
```
mihaelpalfi, line 28: Does this mean that we will always have to add tests imported libraries to the ignore patterns?

mihaelpalfi (approved):
Tests are running again. Good work! Just one thing that gives me second thoughts is this always adding libraries to ignore patterns in the config.

stencil.config.ts
chavlji: I'm asking myself the same thing :( Couldn't we just keep transpile for all js files?

chavlji (approved):
Great job Fermin!

stencil.config.ts
f-jauregui: Not all of them, just the ones that are using ESM.
I could not find other way. We can try this https://nodejs.org/api/vm.html#vm_class_vm_module but it's still in experimental phase, that's why Jest doesn't support it.

chavlji: No problem. It's ok for now. We can try to improve it later. It's great to have tests back! 🚢 it.

f-jauregui: Cool. And just to clarify why Jest doesn't support this we'all can see it here https://jestjs.io/docs/en/ecmascript-modules at the beginning.

PR #80, opened 2021-01-04
BAN-115: Add test cases to tooltip and Notifications

mihaelpalfi:
[JIRA ticket](https://reciprocitylabs.atlassian.net/browse/BAN-115)

#### Description
As per JIRA ticket

#### ChangeLOG

#### Useful links


src/components/zen-tooltip/test/zen-tooltip.spec.tsx
```diff
@@ -1,19 +1,44 @@
...
-        <mock:shadow-root>
-          <slot></slot>
-          <div class="arrow dark top"></div>
-        </mock:shadow-root>
-      </zen-tooltip>
-    `);
+    expect(page.root.getAttribute('variant')).toEqual(variant);
+  });
+
+  it.each(['top', 'right', 'bottom', 'left'])('Test each position %s is displayed correct', async position => {
```
chavlji, line 13: nit: Test name should be `applies corret position attribute is %s`

src/components/zen-tooltip/test/zen-tooltip.spec.tsx
```diff
@@ -1,19 +1,44 @@
...
+    const page = await newSpecPage({
+      components: [ZenTooltip],
+      html: `<zen-tooltip position="${position}" label="Testing tooltip"></zen-tooltip>`,
+    });
+    expect(page.root.getAttribute('position')).toEqual(position);
```
chavlji, line 18: This test only checks if Stencil is applying the correct attribute. We should test our component.

src/components/zen-tooltip/test/zen-tooltip.spec.tsx
```diff
@@ -1,19 +1,44 @@
...
     });
-    expect(page.root).toEqualHtml(`
-      <zen-tooltip class="dark tooltip top" style="max-height: none;">
-        <mock:shadow-root>
-          <slot></slot>
-          <div class="arrow dark top"></div>
-        </mock:shadow-root>
-      </zen-tooltip>
-    `);
+    expect(page.root.getAttribute('variant')).toEqual(variant);
```
chavlji, line 10: Same as for position test. This test isn't very useful.

src/components/helpers/types.ts
```diff
@@ -1,11 +1,14 @@
 // Props:
-
 export type Align = 'left' | 'right' | 'center';
 
 export type Position = 'top' | 'right' | 'bottom' | 'left';
 
 export type TooltipVariant = 'dark' | 'light' | 'error';
 
+export type NotificationVariant = 'success' | 'info' | 'warning' | 'error';
```
chavlji, line 8: Nice!

chavlji (changes requested):

src/components/zen-tooltip/test/zen-tooltip.spec.tsx
mihaelpalfi: Changed

src/components/zen-tooltip/test/zen-tooltip.spec.tsx
mihaelpalfi: You are right. Changed that it checks if correct class is applied.

src/components/zen-tooltip/test/zen-tooltip.spec.tsx
mihaelpalfi: Same as above.

src/components/helpers/types.ts
mihaelpalfi: Still think that here should be an enum :)

src/components/zen-tooltip/test/zen-tooltip.spec.tsx
```diff
@@ -7,16 +7,24 @@ describe('Test parameters rendering', () => {
...
+      const page = await newSpecPage({
+        components: [ZenTooltip],
+        html: `<div>Trigger</div><zen-tooltip position="${position}" always-visible="true" label="Testing tooltip"></zen-tooltip>`,
+      });
+
+      // We have to wait that tooltip is displayed because of always visible param
+      setTimeout(async () => {
+        expect(page.root.classList.contains(position)).toBe(true);
+        done();
+      }, 101);
```
chavlji, line 25: Why 101? Doesn't it work if we lower that to eg. 10? Imo the point is just to postpone it to next cycle.

chavlji (approved):

src/components/helpers/types.ts
chavlji: Agree with that. I'll take another look into that.

src/components/zen-tooltip/test/zen-tooltip.spec.tsx
mihaelpalfi: Because somebody implemented that it should wait 100ms before showing :)

PR #86, opened 2021-01-07
BAN-160: POC for Cypress visual regressions testing

mplesec:
[JIRA ticket](https://reciprocitylabs.atlassian.net/browse/BAN-160)

#### Description
Since Cypress doesn't play nice with iframes (even with the iframe plugin) -> the selector tooling doesn't work (need to manually specify them, which is time consuming).

Running VRT with some stuff in frames causes screenshots to contain other things on the screen besides the queried element (i.e. it sometimes takes the screenshot from 0, 0 of parent viewport to bottom-right of element in iframe that was queried) -> this is super prone to flakiness, thus can't be used this way reliably

A way around this is to go to the URL of the iframe. Though, couldn't find a way to also have story controls in own page the same way as we can click on the icon on the default preview for a component (that only opens that preview). However, manually changing the `viewMode` URL param to `docs` does this (`http://localhost:6006/iframe.html?id=buttons-button--default&viewMode=docs`), and displays all of the content which is in the iframe by default. In the example test, I manually got the id, but we can parse the id out from the anchor `href` attribute anyway.

This way we can effectively test what we get in the iframe in storybook without it actually being an iframe, which means no need for messing around with cypress and its plugins for this stuff.

#### ChangeLOG

#### Useful links


README.md
```diff
@@ -33,10 +33,17 @@ yarn create:component zen-{component-name}
 ```
 
 #### Testing
+
+To run unit and integration tests in Jest:
 ```bash
 yarn test
 ```
 
+To run E2E/integration tests in Cypress:
```
chavlji, line 42: nit: Could also mention it tests a visual regression

chavlji (approved):
Great job! 🏆 

Imo this looks waaay better than Puppeteer (you can see what it's actually been doing).

README.md
mplesec: Will add it to the docs after we make it mandatory on test runs.

f-jauregui (approved):

PR #87, opened 2021-01-08
BAN-159: Playground should support Vue.js

chavlji:
[JIRA ticket](https://reciprocitylabs.atlassian.net/browse/BAN-159)


mplesec (approved):
Awesome 😎 Can be merged as is, since it already provides what's needed. One thing to think about is to provide the minimal necessary functionality to the end-user. What I'm thinking is to only allow for the source area to consume what would be in the template, so like this:
Screenshot 2021-01-08 at 14 54 37
Screenshot 2021-01-08 at 14 54 40

Since this is the way we specify in the docs for consuming a component:
Screenshot 2021-01-08 at 14 56 22

chavlji:
@mplesec I'm merging this for now and we can open a ticket for these improvements.
So you mean we should keep just the plain Html/template thing in the textarea? Keeping only the selected text:
Screenshot 2021-01-11 at 12 44 00
Screenshot 2021-01-11 at 12 43 51

It's surely doable, but how could you then input script stuff like vue data, methods, vanilla js event handlers,...?

mplesec:
@chavlji oh right it needs JS too. Hah, tricky. Keep it like this for now.

PR #108, opened 2021-01-15
PLAT-835: Emmit Native Events

chavlji:
[JIRA ticket](https://reciprocitylabs.atlassian.net/browse/PLAT-835)


mplesec (approved):
Awesome 👍 I also like how the changelog communicates this now. One thing missing around this would be some tests, but these can be added later.

PR #109, opened 2021-01-18
PLAT-839: Fix small glitches

f-jauregui:
[JIRA ticket](https://reciprocitylabs.atlassian.net/browse/PLAT-839)

#### Description
As per JIRA ticket

#### ChangeLOG

#### Useful links

### DIFF
--- a/src/components/zen-button/zen-button.stories.mdx
+++ b/src/components/zen-button/zen-button.stories.mdx
@@ -100,7 +100,7 @@ Button component with slots capabilities
 
 ### 4. With Icons
 
-- Apply leadingIcon and trailingIcon value for the slot attribute in your inner custom HTML
+- Apply leadingIcon and trailingIcon values for the slot attributes in your inner custom HTML
 
 <Canvas>
   <Story name="StoryWithLeadingSlot">{StoryWithIcons.bind({})}</Story>
--- a/src/components/zen-button/zen-button.stories.mdx
+++ b/src/components/zen-button/zen-button.stories.mdx
@@ -1,6 +1,5 @@
 import { Meta, Story, Canvas, ArgsTable } from '@storybook/addon-docs/blocks';
 import { html } from 'lit-html';
-import Readme from './readme.md';
 import { getArgTypes, getDefaultArgs, getComponentData, spreadArgs } from '../../../.storybook/helpers/argTypes';
 
 import data from '../../../stencilDocs.json';
--- a/src/components/zen-button/zen-button.stories.mdx
+++ b/src/components/zen-button/zen-button.stories.mdx
@@ -72,13 +72,7 @@ export const StoryWithIcons = () =>
         <zen-button>
           <i slot="leadingIcon" style="width: 0.625rem;" .innerHTML="${litHtmlIcon(faUser)}"></i>
         </zen-button>
-        <span style="margin-top: 0.5rem;">Just icon on Leading Slot</span>
-      </div>
-      <div style="display: flex; flex-direction: column; align-items: center;">
-        <zen-button>
-          <i slot="trailingIcon" style="width: 0.625rem;" .innerHTML="${litHtmlIcon(faUser)}"></i>
-        </zen-button>
-        <span style="margin-top: 0.5rem;">Just icon on Trailing Slot</span>
+        <span style="margin-top: 0.5rem;">Just icon</span>
       </div>
     </div>
   `;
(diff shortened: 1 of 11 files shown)


chavlji (changes requested):
Looks nice, I'm just missing 2 things from the ticket:

3. should be able to have a button with icon only. Due to some of Jan's latest changes it's now broken (label is always visible and thus it's margin is applied)
4. 
6. Slots description: should be ON the left instead Slot for the icon at the left

src/components/zen-button/zen-button.stories.mdx
```diff
@@ -12,44 +11,76 @@ import { faUser, faChevronDown } from '@fortawesome/pro-light-svg-icons';
...
+        </zen-button>
+        <span style="margin-top: 0.5rem;">Leading and Trailing Slots</span>
+      </div>
+      <div style="display: flex; flex-direction: column; align-items: center;">
+        <zen-button>
+          <i slot="leadingIcon" style="width: 0.625rem;" .innerHTML="${litHtmlIcon(faUser)}"></i>
+        </zen-button>
+        <span style="margin-top: 0.5rem;">Just icon on Leading Slot</span>
+      </div>
+      <div style="display: flex; flex-direction: column; align-items: center;">
```
chavlji, line 77: No need for this story since it looks the same as the icon in the leading slot.

src/components/zen-button/zen-button.stories.mdx
```diff
@@ -12,44 +11,76 @@ import { faUser, faChevronDown } from '@fortawesome/pro-light-svg-icons';
...
+        <zen-button>
+          <i slot="leadingIcon" style="width: 0.625rem;" .innerHTML="${litHtmlIcon(faUser)}"></i>
+          Leading and Trailing Slots
+          <i slot="trailingIcon" style="width: 0.625rem;" .innerHTML="${litHtmlIcon(faChevronDown)}"></i>
+        </zen-button>
+        <span style="margin-top: 0.5rem;">Leading and Trailing Slots</span>
+      </div>
+      <div style="display: flex; flex-direction: column; align-items: center;">
```
chavlji, line 71: In this Figma designs this kind of button should be a perfect square:
Screenshot 2021-01-20 at 15 55 19

chavlji (changes requested):

src/components/zen-button/zen-button.stories.mdx
f-jauregui: So you think that button should modify its padding when the icon is not square?

chavlji: Not sure what would be the best solution here...
I see no other solution but to check if only `leadingIcon` slot exists - has some nodes assigned (and nothing else - no `default` slot, no `trailingIcon` slot) and then apply some class that will take care of it.
We could provide another component called `zen-icon-button`, but imo it's more natural to have just one button component since we already provide `icon` slot in it.

f-jauregui: Why not to have a zen-icon component that will deal with the icon box size?

Then you just go:
```js
<zen-button ...props>
  <zen-icon ...props>
    {ICON_GOES_HERE}
  </zen-icon>
</zen-button>
```

src/components/zen-button/zen-button.stories.mdx
f-jauregui: Done.

src/components/zen-button/zen-button.stories.mdx
chavlji: Remember when I said it would be better to just expose slot `content` instead of `leading`, `trailing` thing? ;)

Imo that the right and most natural way, yeah. To just have a default slot and that's it.

We should think of how we'll control the spacing between icon and label? Designers expect that to be standardized. Might be best to use css sth like:
`:slotted{ .zen-icon + * { margin-left: 1rem; }}`
And app can override that if it needs to.
 
But we need `zen-icon` component anyway. The task is already there for quite some time.

chavlji (approved):
We'll take care of icon-only buttons in the next PR.

src/components/zen-button/zen-button.stories.mdx
f-jauregui: Created zen-icon story in https://reciprocitylabs.atlassian.net/browse/PLAT-935 to tackle this, since it doesn't exist.

PR #112, opened 2021-01-20
Modal windows - POC

chavlji:
POC for `Zen-modal` component.

Check story POC - Modal

Missing a lot of useful props and styles, but it correctly shows the last opened window on top regardless of its position in the dom.

In the end, I thought it would be more appropriate to let the modal's parent controls the opened state through its `show` prop instead of dispatching events. It seems more natural.

A good thing is that everything stays in its place and thus reactive frameworks could easily be applied to modals content.

Global var `window.ZenModalsService` is used to ensure all apps will use the same service when showing modals.

If you're ok with this global service var, the same approach could easily be applied to the notifications stack.
The beauty of this is that consumers don't need to care about what and when to initialize. Nor we need to think about who should own the notifications/modals service.

### DIFF
--- a/src/components.d.ts
+++ b/src/components.d.ts
@@ -177,6 +177,10 @@ export namespace Components {
         "required": false;
     }
     interface ZenModal {
+        /**
+          * Set to true to show and false to hide modal
+         */
+        "show": boolean;
     }
     interface ZenNotification {
         /**
@@ -628,6 +632,10 @@ declare namespace LocalJSX {
         "required"?: false;
     }
     interface ZenModal {
+        /**
+          * Set to true to show and false to hide modal
+         */
+        "show"?: boolean;
     }
     interface ZenNotification {
         /**
--- a/src/components.d.ts
+++ b/src/components.d.ts
@@ -176,6 +176,8 @@ export namespace Components {
          */
         "required": false;
     }
+    interface ZenModal {
+    }
     interface ZenNotification {
         /**
           * Can dismiss
@@ -394,6 +396,12 @@ declare global {
         prototype: HTMLZenLabelElement;
         new (): HTMLZenLabelElement;
     };
+    interface HTMLZenModalElement extends Components.ZenModal, HTMLStencilElement {
+    }
+    var HTMLZenModalElement: {
+        prototype: HTMLZenModalElement;
+        new (): HTMLZenModalElement;
+    };
     interface HTMLZenNotificationElement extends Components.ZenNotification, HTMLStencilElement {
     }
     var HTMLZenNotificationElement: {
@@ -450,6 +458,7 @@ declare global {
         "zen-input": HTMLZenInputElement;
         "zen-input-support-text": HTMLZenInputSupportTextElement;
         "zen-label": HTMLZenLabelElement;
+        "zen-modal": HTMLZenModalElement;
         "zen-notification": HTMLZenNotificationElement;
         "zen-option": HTMLZenOptionElement;
         "zen-spinner": HTMLZenSpinnerElement;
@@ -618,6 +627,8 @@ declare namespace LocalJSX {
          */
         "required"?: false;
     }
+    interface ZenModal {
+    }
     interface ZenNotification {
         /**
           * Can dismiss
@@ -770,6 +781,7 @@ declare namespace LocalJSX {
         "zen-input": ZenInput;
         "zen-input-support-text": ZenInputSupportText;
         "zen-label": ZenLabel;
+        "zen-modal": ZenModal;
         "zen-notification": ZenNotification;
         "zen-option": ZenOption;
         "zen-spinner": ZenSpinner;
@@ -796,6 +808,7 @@ declare module "@stencil/core" {
             "zen-input": LocalJSX.ZenInput & JSXBase.HTMLAttributes<HTMLZenInputElement>;
             "zen-input-support-text": LocalJSX.ZenInputSupportText & JSXBase.HTMLAttributes<HTMLZenInputSupportTextElement>;
             "zen-label": LocalJSX.ZenLabel & JSXBase.HTMLAttributes<HTMLZenLabelElement>;
+            "zen-modal": LocalJSX.ZenModal & JSXBase.HTMLAttributes<HTMLZenModalElement>;
             "zen-notification": LocalJSX.ZenNotification & JSXBase.HTMLAttributes<HTMLZenNotificationElement>;
             "zen-option": LocalJSX.ZenOption & JSXBase.HTMLAttributes<HTMLZenOptionElement>;
             "zen-spinner": LocalJSX.ZenSpinner & JSXBase.HTMLAttributes<HTMLZenSpinnerElement>;
--- a/src/components/zen-modal/zen-modal.scss
+++ b/src/components/zen-modal/zen-modal.scss
@@ -1,0 +1,7 @@
+:host {
+
+}
+
+p {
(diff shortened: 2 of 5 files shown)


mplesec:
Like it, good work 👍

Based on the last meeting we had, I had a thought about this. Let's think through where the separation is between how it looks and how it works.

I think it makes sense to say that the looks are is this:
- ZenModal, ZenModalHeader, ZenModalBody, ZenModalFooter
  - ZenModalHeader contains a default slot for title (and the X button is shown when that default slot is used)

ZenModalHeader and ZenModalFooter would have defaults
- Providing only the default slot would set the title, have the X button on top-right, and have the default Confirm and Cancel buttons in the footer
  `<zen-modal>Are you sure?</zen-modal>`
- To override this, the consumer would use the ZenModalHeader, ZenModalBody, and ZenModalFooter components. Out of these the consumer can then create i.e.:
  ```
  <zen-modal isOpen="isOpen">
    <zen-modal-header close-button=false>
    My Title
    </zen-modal-header>
    <zen-modal-body>
      Simple text.
      or some complex stuff:
      <zen-input></zen-input>
      <zen-button></zen-button>
    </zen-modal-body>
    <zen-modal-footer>
      <zen-button>No.</zen-button>
      <zen-button onClick=showConfirmation>Go time</zen-button>
    </zen-modal-footer>
    
    // For nested
    <zen-modal isOpen="nestedIsOpen">
    </zen-modal>
  </zen-modal>
  ```

For events, the default buttons (and X button) fire some default events on ZenModal, i.e. onConfirmed and onCancelled. When the consumer specifies their own buttons they can then handle their own events. Important to note is that in both cases the consumer handles when the modal is shown!

Regarding nesting, if we could avoid having a dedicated service (which seems to be a global modal state store) for this all the better. It's a different case than with notifications since those do not have any markup and are only a string event fired (and also don't require coms back to the consumer). Here the markup for a modal must be specified and it seems like having an extra global singleton service for what is only presentational would encourage consumers to not follow web standards and probably start relying on events too much which would lead to some weird race condition bug down the line.

In general, might make sense to just start with the most basic modal and add a store and event bus later if we encounter some case where we absolutely need it. 😅

chavlji:
So `zen-modal-header`, `zen-modal-body` would be a new fragment component? Would it make sense to just expose slots on `zen-modal` where `zen-body` would be a default slot?
```<zen-modal>
  <div slot="header">
    <custom-icon /> Risk editor
  </div>
  Here you can edit risk:
  <input id="input1" />
  <input id="input2" />
  <zen-button onClick="this.openFormulaEditor()">Edit formula</zen-button>
  <div slot="buttons">
    <zen-button onClick="this.cancel()">Cancel</zen-button>
    <zen-button onClick="this.apply()">Apply</zen-button>
  </div>
</zen-modal>```

This usage isn't enough:
`<zen-modal>Are you sure?</zen-modal>`
We wouldn't know what's the title, which buttons to provide (just 'ok' or also 'cancel'). We'd need to support all those things through props which means duplicating things you can achieve through slots.

In the case of custom buttons, consumer would need to take care of closing the modal, right? Is that what you've meant with  *"consumer handles when the modal is shown"*?

I don't quite understand what you mean with *" having an extra global singleton service for what is only presentational would encourage consumers to not follow web standards and probably start relying on events too much which would lead to some weird race condition bug down the line"*? Is that meant for using the same singleton approach with notifications? I guess we can assume that notifications will never need a content slot? I mean we can always provide some additional variants, right? If that's true I'd also rather go with events (seems more natural). Though we'd still need some singleton service that will handle those events and create and show <zen-notification> elements? What do you mean by *"consumers 
 would start relying on events too much"*?

**Comment by mplesec** [comment]
Was just as an example. Slots would work as well, we don't need fragment components in this case.

The simple case would mean that some default would be present. In this case: `<zen-modal>Are you sure?</zen-modal>` the default would be the two buttons in the footer, the X button in the header, and the `Are you sure?` text would serve as the title.

Regarding the events, the component emits them and it's up to the consumer to handle them (which means setting a variable that closes the modal.

My point of view is that the service isn't (at least yet) needed for the modals. It is needed for the notifications.

**Comment by chavlji** [comment]
Agree.
@f-jauregui do you have anything to add?

**Comment by f-jauregui** [comment]
I think that this way to structure the modal is good.
As far is I can see (joining with previous comments) is (correct me if i'm wrong):
```js
<zen-modal>
  <div slot="header" />
  { content that goes in the default slot }
  <div slot="footer" />
</zen-modal>
```
So if we don't fulfill header or footer slot we just render the default ones (Let's say close button for header; accept and cancel buttons for the footer).
In that way we can call:
```js
<zen-modal>
  Are you sure about that?
</zen-modal>
```
And we show a modal with the question as content. Default content for header and footer.
If we want something special for header or footer we just use the slots.

I'm more concerned about how this component will be triggered, I don't think that we need the modals service. If we can rely on `position: fixed` and [render tree construction](https://developers.google.com/web/fundamentals/performance/critical-rendering-path/render-tree-construction) that will make it simpler and easier to dynamically modify modal content from the consumer side.

Would be great if we could see something from the consumer perspective, but so far I liked it.

chavlji:
@f-jauregui Your examples are correct. I was a bit skeptic about the second one without slots, but I guess we can render title prop in header and `OK` button in the footer if it's not provided through slots.

Service is mandatory for two reasons:
1. We want dimmer to be shown only for the topmost window
2. Get a propper layering

Render tree construction is new to me. So we have CSSOM as some sort of detached nodes and when we want to render it, we "insert" it into DOM? How would you know where you need to "insert" your modal? Can you explain it a bit more?

f-jauregui:
> So we have CSSOM as some sort of detached nodes and when we want to render it, we "insert" it into DOM?

Yes, you [append](https://developer.mozilla.org/en-US/docs/Web/API/ParentNode/append) it to some DOM parent. As you append the component, this will be added last and by tree construction rendering will be made at the top.

> How would you know where you need to "insert" your modal?

Triggered by events like a click in some button.

> Can you explain it a bit more?

Let's go through this with a sync.

mplesec:
Maybe I misunderstood the above. What I was thinking as the simplest implementation is to rely only on the DOM tree as-in, no moving around node's necessary. The modal's node wouldn't get appended anywhere, that's the key. It's in the DOM tree where the consumer defines it to be. It gets shown when the conditional specifies so.
```
/* Defined in a component's DOM by the consumer */

// Modal is defined in the DOM, but hidden
<div>
  <button @click="toggleModal">Click me</button>
  <zen-modal *if="isShown"></zen-modal> // isShown = false
</div>

// Modal is shown, and the DOM tree remains unchanged
<div>
  <button @click="toggleModal">Click me</button>
  <zen-modal *if="isShown"></zen-modal>  // isShown = true
</div>
```

Reason why I'd want to go with the simplest approach first is because adding a modal service later would easily build on this. No cost. And it saves us time initially.

chavlji:
I'm also against moving elements around the DOM. It's not our business and we can easily break some framework connections by doing this. Plus I can't see any practical benefits.

@f-jauregui why don't you like the modal service? For me it seems useful approach for all components that needs some kind of central management.

I agree though we should make it as simplest as possible.

If we agree that:
1. ...modal is rendered and controlled by the app
2. ...when modal is shown is controlled by the app through prop `show`
...then I guess we have I/O contract on which we can build and we can change/improve internal stuff later. Without introducing breaking changes that is.

chavlji:
We'll continue here:
https://github.com/reciprocity/zen-ui/pull/133

PR #113, opened 2021-01-20
BAN-164: Fix existing and improve incomplete tests

f-jauregui:
Updated tests for:

- input
- checkbox
- button
- textarea
- spinner
- label
- support text

[JIRA Ticket - Fix existing tests](https://reciprocitylabs.atlassian.net/browse/BAN-164)
[JIRA Ticket - Improve incomplete tests](https://reciprocitylabs.atlassian.net/browse/BAN-165)

#### Description
As per JIRA ticket

#### ChangeLOG

#### Useful links

### DIFF
--- a/src/components/zen-button/test/zen-button.spec.tsx
+++ b/src/components/zen-button/test/zen-button.spec.tsx
@@ -58,4 +58,23 @@ describe('zen-button variants rendering', () => {
     expect(page.root.querySelector('[slot="leadingIcon"]')).toBeTruthy();
     expect(page.root.querySelector('[slot="trailingIcon"]')).toBeTruthy();
   });
+
+  it('triggers click event on onKeyDown Enter key', async () => {
+    const page = await newSpecPage({
+      components: [ZenButton],
+      html: `<zen-button></zen-button>`,
+    });
+
+    const keyupSpy = jest.fn();
+    const clickSpy = jest.fn();
+
+    page.root.addEventListener('keyup', keyupSpy);
+    page.root.addEventListener('click', clickSpy);
+    page.root.dispatchEvent(new KeyboardEvent('keyup', { code: 'Enter' }));
+
+    await page.waitForChanges();
+
+    expect(keyupSpy).toHaveBeenCalled();
+    expect(clickSpy).toHaveBeenCalled();
+  });
 });
--- a/src/components/zen-button/test/zen-button.spec.tsx
+++ b/src/components/zen-button/test/zen-button.spec.tsx
@@ -10,54 +10,18 @@ describe('zen-button variants rendering', () => {
     expect(page.root.shadowRoot).toBeTruthy();
   });
 
-  it.each(['primary', 'secondary', 'tertiary', 'positive', 'destructive'])('renders %s button', async variant => {
-    const page = await newSpecPage({
-      components: [ZenButton],
-      html: `<zen-button variant="${variant}"></zen-button>`,
-    });
-    expect(page.root.classList.contains(`btn-${variant}`)).toBe(true);
-  });
-
-  it('renders with loading state', async () => {
-    const page = await newSpecPage({
-      components: [ZenButton],
-      html: `<zen-button loading></zen-button>`,
-    });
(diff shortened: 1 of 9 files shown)


mihaelpalfi (approved):
Now it looks like unit tests. Great job Fermin!

mplesec (approved):
Screenshot 2021-01-20 at 19 09 02

src/components/zen-checkbox/test/zen-checkbox.spec.tsx
```diff
@@ -2,116 +2,116 @@ import { newSpecPage } from '@stencil/core/testing';
...
     });
-    expect(page.root).toEqualHtml(`
-      <zen-checkbox checked="" disabled="">
-        <mock:shadow-root>
-          <input checked="" class="input-control" type="checkbox" disabled="">
-          <label class="disabled"></label>
-        </mock:shadow-root>
-      </zen-checkbox>
-    `);
+    expect(page.root.getAttribute('disabled')).toBe('');
```
chavlji, line 48: These tests aren't really testing much? You've setting the attr in html and then check that element has this attribute? Why wouldn't it?

Imo we should have this kind of tests just for one initial render...

src/components/zen-button/test/zen-button.spec.tsx
```diff
@@ -1,172 +1,80 @@
...
-        <mock:shadow-root>
-          <slot name="leadingIcon"></slot>
-          <span class="ml"><slot>Button</slot></span>
-          <slot name="trailingIcon"></slot>
-        </mock:shadow-root>
-        Button
-        <zen-spinner slot="leadingIcon"></zen-spinner>
-      </zen-button>
-    `);
+    expect(page.root.querySelector('[slot="leadingIcon"]')).toBeTruthy();
```
chavlji, line 42: As mentioned. We're just checking if div has things we've set in the template string...

chavlji (changes requested):
I'm not sure. A lot of these tests just set HTML through a template string and then test if dom has things we've set in the template string.

They are just testing that converting a template string into HTML works, which means it's not a component unit test.

src/components/zen-button/test/zen-button.spec.tsx
f-jauregui: This checks if an element is rendered outside the shadow dom with attribute slot set to leadingIcon. Should I remove this?

src/components/zen-checkbox/test/zen-checkbox.spec.tsx
f-jauregui: This check if props are being properly applied to the input. Should I remove it?

chavlji: If you set it in the HTML template string, how can it not be set? Only if the parser fails, but this isn't our responsibility.
I'd remove all such tests because I can't see any value in them.

We've had a discussion with @mplesec that it might only be reasonable to test whether all expected props are still supported (so that we don't accidentally delete some prop). But they should all be tested in one test. Possibly using some helper function like:
``` expect(reflectsPropsOnAttributes(['value', 'disabled', 'variant']). toBeTruthy()```

src/components/zen-button/test/zen-button.spec.tsx
f-jauregui: Removed it.

src/components/zen-checkbox/test/zen-checkbox.spec.tsx
f-jauregui: Done.

And please include me when this happens:

> We've had a discussion with @mplesec that it might only be reasonable to test whether all expected props are still supported (so that we don't accidentally delete some prop). But they should all be tested in one test. Possibly using some helper function like:
expect(reflectsPropsOnAttributes(['value', 'disabled', 'variant']). toBeTruthy()

chavlji: Sure. We've had this discussion just today in the morning, so after this pr...

chavlji (approved):
👍🏻

PR #117, opened 2021-01-21
PLAT-854: Add avatar component

mihaelpalfi:
[JIRA ticket](https://reciprocitylabs.atlassian.net/browse/PLAT-854)

#### Description
As per JIRA ticket

#### ChangeLOG

#### Useful links


f-jauregui (commented):
So far it looks really good. Nice job, Mihael!
Waiting for the tests now! 😎

src/components/zen-avatar/zen-avatar.tsx
```diff
@@ -0,0 +1,37 @@
...
+  @Prop() readonly imageUrl: string = '';
+
+  /** Background color  */
+  @Prop() readonly background: string = '#abd2f5';
+
+  /** Font color  */
+  @Prop() readonly color: string = '#00528c';
+
+  /** Name Initials  */
+  @Prop() readonly label: string = '';
```
chavlji, line 20: This prop should be `userName` instead. Component should provide the logic to extract the initials. Otherwise each app will need to implement it on it's own...
Sth like: `userName.match(/(\b([A-Z]|[a-z]))/g).join('').substring(0,2).toUpperCase()`

src/components/zen-avatar/zen-avatar.tsx
```diff
@@ -0,0 +1,37 @@
...
+  @Prop() readonly size: AvatarSize = 'large';
+
+  hasImage(): boolean {
+    return this.imageUrl != '';
+  }
+
+  render() {
+    return (
+      <Host style={{ background: this.background, color: this.color }} class={this.size}>
+        <img class={{ hidden: !this.hasImage(), [this.size]: true }} src={this.imageUrl} />
```
chavlji, line 32: No need for size classes. Not here nor on host element.
Use attribute selectors in the css instead:
`:host([size=md])` and `:host([size=md]) > img`
Also remove tests where you test if size classes are applied.

src/zen-styles/mixins/helpers.scss
```diff
@@ -7,3 +7,13 @@
   user-select: none; /* Non-prefixed version, currently
                                   supported by Chrome, Edge, Opera and Firefox */
 }
+
+@mixin round-icon($size, $padding: null) {
+  margin: $padding;
```
chavlji, line 12: Why not just setting `width` and `height` instead of both `min/max-width`?

src/zen-styles/mixins/helpers.scss
```diff
@@ -7,3 +7,13 @@
   user-select: none; /* Non-prefixed version, currently
                                   supported by Chrome, Edge, Opera and Firefox */
 }
+
+@mixin round-icon($size, $padding: null) {
```
chavlji, line 11: Move this mixin into `zen-avatar.scss` because it's very specific and not widely reusable.

chavlji (changes requested):
Very nice! Looks great.

src/components/helpers/types.ts
```diff
@@ -9,6 +9,8 @@ export type NotificationVariant = 'success' | 'info' | 'warning' | 'error';
 
 export type Duration = 'none' | 'short' | 'medium' | 'long';
 
+export type AvatarSize = 'medium' | 'large';
```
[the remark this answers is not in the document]
mihaelpalfi: Changed.

src/components/zen-avatar/zen-avatar.tsx
mihaelpalfi: Done

src/components/zen-avatar/zen-avatar.tsx
mihaelpalfi: Good one. Didnt know this is possible. Changed.

src/zen-styles/mixins/helpers.scss
mihaelpalfi: My opinion this is a helper mixin. It should be here if there is a need for it in the future. When its not somebody will put it there again and not remove the one from avatar component.

src/zen-styles/mixins/helpers.scss
mihaelpalfi: Changed.

src/components/zen-avatar/zen-avatar.tsx
```diff
@@ -17,20 +17,32 @@ export class ZenAvatar {
...
+    let initials = '';
+    if (this.userName) {
+      initials = this.userName
+        .match(/(\b([A-Z]|[a-z]))/g)
+        .join('')
+        .substring(0, 2)
+        .toUpperCase();
+    }
+    return initials;
+  }
```
mplesec, line 39: This assumes a username will have a space in? Also, please add some tests for this function.

chavlji: It gets words using word boundary. So can be any character that breaks words.
Agree that tests should be added.

src/zen-styles/mixins/helpers.scss
chavlji: Ok. Maybe we should just use some different name for this mixin. Its content doesn't have any icon... `roundel`, `round-icon-wrapper` maybe? It should probably also need to take bg-color as a parameter, to be really useful?

chavlji (approved):

src/components/zen-avatar/zen-avatar.tsx
mihaelpalfi: Added additional case that takes initials from email when username dosent exist. Also if no surname is provided. Added also tests fro all this cases.

src/components/zen-avatar/zen-avatar.tsx
```diff
@@ -29,11 +32,19 @@ export class ZenAvatar {
...
+          .match(/(\b([A-Z]|[a-z]))/g)
+          .join('')
+          .substring(0, 2)
+          .toUpperCase();
+      } else {
+        // Get initials oly from name
+        initials = this.userName.substring(0, 2).toUpperCase();
+      }
+    } else {
+      initials = this.email.substring(0, 2).toUpperCase();
```
chavlji, line 47: I'd expect that `jan.savli@gmail.com` to result in `JA`. I would expect `JS`.
To achieve this you can use the same function as for userName. Same regex...

chavlji (approved):
Nice!

Would be good to change those two things, but you can do it on the next ticket if you want.

Please if you can just add `noselect` mixin and `cursor: default`, so the initials can't be selected ...and we can merge it right away.

src/components/zen-avatar/zen-avatar.tsx
mihaelpalfi: Great suggestion. But requirement is that it should take first two letters of email.

PR #122, opened 2021-01-25
PLAT-852: Zen Table

mplesec:
[JIRA ticket](https://reciprocitylabs.atlassian.net/browse/PLAT-852)

#### Description
~~Quick POC for the Zen Table components. Comment if you'd expect anything different, especially from the consumer's side.~~
Basic components for `ZenTable`.

#### ChangeLOG

#### Useful links


src/components/zen-table-cell/zen-table-cell.tsx
```diff
@@ -0,0 +1,18 @@
...
+@Component({
+  tag: 'zen-table-cell',
+  styleUrl: 'zen-table-cell.css',
+  shadow: true,
+})
+export class ZenTableCell {
+  render(): HTMLElement {
+    return (
+      <Host>
+        <td>
```
chavlji, line 12: Would it be possible to get rid of this wrapping table elements into host each time? So that Host itself becomes a td? It might be possible to set just the right css (display: table,...) and avoid tables altogether?

Is this syntax even correct? Shouldn't eg. `<tr>` always contain just `<td>` elements? In our case it contains `<zen-table-cell>` custom elements which isn't `<td>`:
```html
<table>
  <some-custom-element>
    <tr>
      <some-custom-element>
        <td />
      </some-custom-element>
      <some-custom-element>
        <td />
      </some-custom-element>
    </tr>
  </some-custom-element>
</table>
```

chavlji (commented):
It looks ok in general.

One thing to consider:
- Component names are maybe too long and look too verbose in the HTML. What about `<zen-tr>` `<zen-td>` `<zen-th>`

src/components/zen-table-cell/zen-table-cell.tsx
mplesec: Possible, check the latest commit. Note however that there are no `tr`, `th`, and `td` elements in the markup then (not sure if that's ok).

leferreyra: I think @chavlji is right about table elements only allowing for certain children types, see the [permitted content section](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/table)

leferreyra (commented):
I like the direction this is going, great work.

@chavlji I like long verbose names :P, these look great IMHO

src/components/zen-table-cell/zen-table-cell.tsx
mplesec: I think the only HTML valid approach then is to wrap a regular table element within a `ZenTable` custom element. The consumer would then add the markup in the same way as for a regular table. Like so:
```
<zen-table>
  <thead>
    <tr>
      <th>Foo</th>
      <th>Bar</th>
    </tr>
  </head>
  <tbody>
    <tr>
      <td>A</td>
      <td>1</td>
    </tr>
    <tr>
      <td>B</td>
      <td>2</td>
    </tr>
  </tbody>
</zen-table>
```

leferreyra: I don't know if this is possible but rendering divs with the correct display could work, maybe worth looking into.

chavlji:
To me, this new commit looks great!
It also behaves as you'd expect it from the table:
Screenshot 2021-01-25 at 14 16 01

src/components/zen-table-cell/zen-table-cell.tsx
mplesec: That's how it works now. Was only mentioning it because i.e. if you run this through the HTML validator the markup won't pass since there are custom elements within the table native elements.

mplesec: @chavlji Though now I'm questioning why even have custom elements instead of these native elements. What's the value for us using that, instead of only having a custom table element wrap the native markup...

chavlji: True! 😄 

Some custom styling might be the only good reason. So Zen-UI defines table-border, row bottom border,...

mplesec: The custom styling would be part of the `ZenTable` custom component.

chavlji: So you would end up with sth like this:
```
<zen-table>
  <tr>
      <td>
        <zen-status-presenter>
      </td>
      <td>
        <zen-title-presenter>
      </td>
  </tr>
</zen-table>
```
To me it looks great!

leferreyra: I would say the reason is Control. The same reason why you don't just give a `?query=` param on an API so the user can input custom SQL in there, you want to control how the user access data. In this case, you want to control how the user uses the table component and its children.

mplesec: Makes sense yeah. This way we enforce that the only way to use this component is with the following custom components:
```
<zen-table>
  <zen-table-row slot="header">
  </zen-table-row>
  <zen-table-row>
  </zen-table-row>
  ...
</zen-table>
```
And we build whatever we build in the component.

mplesec: Screenshot 2021-01-25 at 15 01 48

mplesec:
Updated what we discussed and added some basics styles, looks like this now:
Screenshot 2021-01-25 at 15 43 17

f-jauregui (approved):
Good work!

src/components/zen-table/zen-table.tsx
```diff
@@ -0,0 +1,28 @@
+import { h, Component, Host, Prop } from '@stencil/core';
+
+@Component({
+  tag: 'zen-table',
+  styleUrl: 'zen-table.scss',
+  shadow: true,
+})
+export class ZenTable {
+  /** Expands table to fill full available width */
+  @Prop() readonly fullWidth: boolean = false;
```
chavlji, line 10: When you wouldn't want that? You always want the host to take up the whole available space?
Besides you can always override that by appending styles to `<zen-table>`.

src/components/zen-table/zen-table.stories.mdx
```diff
@@ -0,0 +1,225 @@
...
+import data from '../../../stencilDocs.json';
+const compData = data.components.find(n => n.tag === 'zen-table');
+const argTypes = getArgTypes(compData);
+
+<Meta title="In progress/Tables/Table" component="zen-table" argTypes={argTypes} />
+
+export const Story1 = () => {
+  return html/*html*/ `
+    <zen-table id="table1">
+      <zen-table-row slot="header">
```
chavlji, line 15: I really don't like these long names... it totally blurs the point. It will get even worse when you have some additional attrs on these elements:

Screenshot 2021-01-26 at 15 32 09

I know that usually won't look that bad due to the framework's loops, but still...

src/components/zen-table-row/zen-table-row.css
```diff
@@ -0,0 +1,3 @@
+:host {
```
chavlji, line 1: This should be a scss file?

chavlji (approved):
Great job! 👍🏻

src/components/zen-table/zen-table.tsx
f-jauregui: > When you wouldn't want that? You always want the host to take up the whole available space?

Agree. And a container element defines box size.

> Besides you can always override that by appending styles to <zen-table>.

If I'm not wrong, we talked about this several times. This is a pattern that we want to avoid, whenever possible.

mplesec: There are use cases where you want it to take up the whole space and adapt the cell heights accordingly, and then you'd want the table to take up as much space as necessary, i.e. when resizing the columns.

src/components/zen-table-row/zen-table-row.css
mplesec: Oh yeah, I needed a variable before.

src/components/zen-table/zen-table.tsx
chavlji: Yeah, but last time we've said we won't do layout components. Bottom line: layout is on the app's side and component shouldn't have to do anything with that.
If you can do it some other way than using styles it's fine...

src/components/zen-table/zen-table.stories.mdx
mplesec: It's a matter of preference I guess. We asked the consumers, @leferreyra expressed a preference for the full name.

leferreyra: full names all day baby

src/components/zen-table-header-cell/test/zen-table-header-cell.spec.tsx
```diff
@@ -0,0 +1,8 @@
+// import { newSpecPage } from '@stencil/core/testing';
+// import { ZenTableCell } from '../zen-table-header-cell';
+
+describe('zen-table-header-cell', () => {
+  it('renders', async () => {
+    // TODO:
+  });
+});
```
f-jauregui, line 8: Tests will be added in this PR?

f-jauregui (changes requested):

src/components/zen-table-header-cell/test/zen-table-header-cell.spec.tsx
mplesec: Later, I'd like to finish the minimum components and props first. I'll be adding the stories though (might just add VRT as well with that).

src/components/zen-table/zen-table.tsx
mplesec: Ah ok. So it would make sense then to use `width: 100%` as the default?

chavlji: Yep. It seems the most useful. If the app needs some other layout, it can change it without a prop.

src/components/zen-table-row/zen-table-row.css
chavlji: So it's standardized. Are there any cons against using css in scss files?

mplesec: None, didn't know this was a rule. I assumed since the scaffolding creates a css file that we use scss only when we need some functionality of it.

src/components/zen-table/zen-table.tsx
mplesec: Makes sense, removed the prop

f-jauregui (approved):

PR #126, opened 2021-01-26
QUICKFIX: Fix flaky Zen-Text VRT

chavlji:
We still need an async solution in this particular case, but it works even with a timeout of 1ms. And that should fix Cypress's problems. I tried to run it 5 times and it passed.

The cause of why the text-details component doesn't pick up the right font-sizes in a sync flow is the following:
- attributes are appended from the start (eg. `size="md"`). I checked that by console.log them
- CSS styles are for some reason applied after the initial render. That means the following:
 - on the first render font-sizes are wrong because the text hasn't been styled yet even though it has the right attributes
 - when components are styled, transitions will be run. That means initial transitions that we don't want from the UX perspective as well as from the testing perspective.

I'll try to take a look to if we can enforce component styles to be there on the initial render.


mplesec (approved):
Awesome 👍

PR #130, opened 2021-01-28
PLAT-1013: ZenTable unit tests

mplesec:
[JIRA ticket](https://reciprocitylabs.atlassian.net/browse/PLAT-1013)

#### Description
Some simple unit tests for these components since there's not much else to test yet.

#### ChangeLOG

#### Useful links

### DIFF
--- a/src/components/zen-table-cell/test/zen-table-cell.spec.tsx
+++ b/src/components/zen-table-cell/test/zen-table-cell.spec.tsx
@@ -1,8 +1,21 @@
-// import { newSpecPage } from '@stencil/core/testing';
-// import { ZenTableCell } from '../zen-table-cell';
+import { newSpecPage } from '@stencil/core/testing';
+import { ZenTableCell } from '../zen-table-cell';
 
 describe('zen-table-cell', () => {
   it('renders', async () => {
-    // TODO:
+    const page = await newSpecPage({
+      components: [ZenTableCell],
+      html: `<zen-table-cell>Content</zen-table-cell>`,
+    });
+
+    expect(page.root).toEqualHtml(`
+      <zen-table-cell>
+        <mock:shadow-root>
+          <slot>
+          </slot>
+        </mock:shadow-root>
+        Content
+      </zen-table-cell>
+    `);
   });
 });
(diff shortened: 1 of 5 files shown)


src/components/zen-table-cell/test/zen-table-cell.spec.tsx
```diff
@@ -1,8 +1,21 @@
...
 
 describe('zen-table-cell', () => {
   it('renders', async () => {
-    // TODO:
+    const page = await newSpecPage({
+      components: [ZenTableCell],
+      html: `<zen-table-cell>Content</zen-table-cell>`,
+    });
+
+    expect(page.root).toEqualHtml(`
```
chavlji, line 11: I'm not sure regarding these snapshot tests... they're not very useful and are very hard to maintain.
We've deleted them from all other components.

chavlji (changes requested):

src/components/zen-table-cell/test/zen-table-cell.spec.tsx
mplesec: I don't know about other component, but in these components they're necessary to test that the slot content is passed in as expected, and to check that an attribute was applied correctly. AFAIK there's no other way to determine correctness than this (apart from visual tests, but those are mean't to mostly test how components look).

chavlji: Aha ok. Agree to that. I think @f-jauregui also had a case where he wanted to check if slot gets rendered.

chavlji (approved):

src/components/zen-table-cell/test/zen-table-cell.spec.tsx
f-jauregui: It's merged, but 🤷‍♂️ 

Shadow DOM of the component can be accessed by the `page.root.shadowRoot` object. 

To check if the slot was rendered:

```js
expect(page.root.shadowRoot.querySelector('slot')).toBeTruthy();
```

To check if some content was applied to a named slot:

```js
expect(page.root.querySelector('[slot="namedSlot"]')).toBeTruthy();
```

And so on...

mplesec: Addressed in https://github.com/reciprocity/zen-ui/pull/135

PR #133, opened 2021-01-29
PLAT-676: Implement modals

chavlji:
[JIRA ticket](https://reciprocitylabs.atlassian.net/browse/PLAT-676)

1. Design is copied from ZenGRC, because we don't have final designs yet.
2. Based on this POC https://github.com/reciprocity/zen-ui/pull/112
3. VRT tests aren't ok (snapshot contains just buttons 😅 ), but this will be addressed in different ticket https://reciprocitylabs.atlassian.net/browse/PLAT-1022


.storybook/helpers/argTypes.js
```diff
@@ -1,6 +1,7 @@
 import stencilDocs from '../../stencilDocs.json';
 import { spread } from '@open-wc/lit-helpers';
 import { camelKeysToKebab } from './utils';
+import { camelCase } from 'lodash';
```
f-jauregui, line 4: Change this import to:
```js
import camelCase from 'lodash/camelCase';
```

f-jauregui (approved):
Great job, Jan! 💯 
Really like the fact that you can edit props when you have the modal open in storybook.

Just that little update in the lodash import and we're all set.

src/components/zen-animate/test/zen-animate.spec.tsx
```diff
@@ -17,6 +18,7 @@ describe('zen-animate', () => {
       components: [ZenAnimate],
       html: `<zen-animate><h1>Slot</h1></zen-animate>`,
     });
+    jest.clearAllTimers();
```
mplesec, line 21: Why is this needed only in this test case?

src/components/zen-animate/zen-animate.tsx
```diff
@@ -24,21 +24,33 @@ export class ZenAnimate {
...
 
     if (!this.show) {
       // Remove element with delay, so transition finishes first:
-      const transition = parseFloat(getComputedStyle(slot)['transitionDuration']) * 1000;
+      const transition = parseFloat(getComputedStyle(slot[0])['transitionDuration']) * 1000;
+
       this.hideTimer = setTimeout(() => {
         this.doShow = false;
-      }, transition);
+      }, transition || 10);
```
mplesec, line 53: Why this default? Why not instant?

src/components/zen-modal/test/zen-modal.spec.tsx
```diff
@@ -0,0 +1,56 @@
...
+  });
+
+  it('emmits OK event with default OK button', async () => {
+    await testClickEvent(page, '.buttons-row .btn-ok', 'okClicked');
+  });
+
+  it('calls makeTopmost on show', async () => {
+    page.root.show = true;
+    await page.waitForChanges();
+    expect(modalsService.makeTopmost).toHaveBeenCalled();
```
mplesec, line 46: A good idea with this is to also check that it was called exactly once.

src/components/zen-modal/test/zen-modals-service.spec.tsx
```diff
@@ -0,0 +1,53 @@
...
+  it('promotes modal to topmost', async () => {
+    modalsService.makeTopmost(modal1);
+    modalsService.makeTopmost(modal2);
+    expect(modalsService.modals.length).toEqual(2);
+    expect(modalsService.modals[1].getAttribute('id')).toEqual('modal2');
+    expect(modal2.getAttribute('topmost')).toEqual('true');
+    expect(modal1.getAttribute('topmost')).toBeFalsy();
+  });
+
+  it('increases lastZIndex', async () => {
```
mplesec, line 38: Semi-related, how problematic could this zindex stuff get if tooltips also have them?

src/zen-styles/variables.scss
```diff
@@ -1,6 +1,7 @@
 @import './colors';
 @import './mixins/index';
 
+$text-2xl: 1.5rem !default;
```
mplesec, line 4: This is per designs? Seems strange that increases by so much after `1.125`.

src/components/zen-modal/zen-modal.stories.mdx
```diff
@@ -0,0 +1,118 @@
...
+
+### 1. Stacking
+
+Modals can be stacked. Last one showed will be the top-most.
+
+<Canvas>
+  <Story name="mutiple">{StoryMultipleModals.bind({})}</Story>
+</Canvas>
+
+## Properties
```
mplesec, line 102: Maybe it would be a good idea to have stories for two components: `Modal` and `Modal window`. Since toggling these doesn't reflect changes until clicking on the buttons.

Also, toggling the `show` property produces a bug:
Screenshot 2021-02-02 at 09 08 21

src/components/zen-modal/zen-modal.stories.mdx
```diff
@@ -0,0 +1,118 @@
...
+import { action } from '../../../.storybook/helpers/custom-action';
+import { getArgTypes, getDefaultArgs, getComponentData, spreadArgs } from '../../../.storybook/helpers/argTypes';
+
+import data from '../../../stencilDocs.json';
+const compData = data.components.find(n => n.tag === 'zen-modal');
+const argTypes = getArgTypes(compData);
+
+<Meta title="Notifications/Modal window" component="zen-modal" argTypes={argTypes} />
+
+export const StoryWithControls = args => {
```
mplesec, line 12: Would be cool to have some of this as instructions in the story description.

src/components/zen-modal/zen-modal.tsx
```diff
@@ -0,0 +1,96 @@
...
+  @Element() hostElement: HTMLZenModalElement;
+
+  /** Set `true` to show and `false` to hide modal */
+  @Prop({ reflect: true }) readonly show = false;
+
+  /** Modal title (irrelevant if slot `header` passed) */
+  @Prop() readonly label: string = 'Zen-UI modal window';
+
+  /** Hide default top-right X and default Cancel button */
+  @Prop() readonly hideCancel: boolean = false;
```
mplesec, line 25: NIT: What if I want only the default buttons but no X?

src/components/zen-modal/zen-modal.tsx
```diff
@@ -0,0 +1,96 @@
...
+
+  /** Top-right X button or default Cancel button clicked */
+  @Event() cancelClicked: EventEmitter<undefined>;
+
+  /** Default Ok button clicked (irrelevant if slot `buttons` passed) */
+  @Event() okClicked: EventEmitter<undefined>;
```
mplesec, line 31: Should these types be `void`?

mplesec (commented):
Good stuff 👍 In the future, please split PRs a bit, this took way longer to review than expected.

src/components/zen-modal/zen-modal.stories.mdx
```diff
@@ -0,0 +1,118 @@
...
+
+# Modal
+
+Standard modal window
+
+## Component previews
+
+<br />
+
+### 1. Stacking
```
mplesec, line 94: Found a strange bug here. Clicking on the X doesn't work with the first modal.

https://user-images.githubusercontent.com/9392804/106571251-3616c280-6537-11eb-98b5-a657616d6c5c.mov

src/components.d.ts
```diff
@@ -749,6 +770,28 @@ declare namespace LocalJSX {
...
+         */
+        "hideCancel"?: boolean;
+        /**
+          * Modal title (irrelevant if slot `header` passed)
+         */
+        "label"?: string;
+        /**
+          * Top-right X button or default Cancel button clicked
+         */
+        "onCancelClicked"?: (event: CustomEvent<undefined>) => void;
```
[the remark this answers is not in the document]
chavlji: Stencil complains:
Screenshot 2021-02-02 at 12 18 01
Renaming it to `cancel` and `confim`

src/components/helpers/helpers.ts
```diff
@@ -61,3 +61,11 @@ export function oppositePosition(position: Position): Position {
 export function getComposedPath(event: MouseEvent): EventTarget[] {
   return event.composedPath();
 }
+
+export function toggleAttribute(element: Element, attribute: string, value: string): void {
```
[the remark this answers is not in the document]
chavlji: This method doesn't seem to allow setting the attribute value which I need in this case (it only appends or removes the attribute).

src/components/zen-animate/test/zen-animate.spec.tsx
```diff
@@ -38,4 +40,25 @@ describe('zen-animate', () => {
...
   });
+
+  it('hides after close transition', async () => {
+    const page = await newSpecPage({
+      components: [ZenAnimate],
+      html: `<zen-animate><h1>Slot</h1></zen-animate>`,
+    });
+    await page.waitForChanges();
+    expect(page.root.shadowRoot.querySelector('slot')).toBeFalsy();
+    // Show it:
```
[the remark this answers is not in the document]
chavlji: You're right. Since next line is `page.rootInstance.show = true;` it's a lot self explanatory.

src/components/zen-animate/test/zen-animate.spec.tsx
chavlji: Not entirely sure... Without this it seems like timer set in `componentDidRender()` of this test is triggered in the 3rd test and it raises an error: `Cannot read property '$hostElement$' of undefined`

src/zen-styles/variables.scss
chavlji: It's not. Designers specified it to max `1.125rem`. I'll change it to `$text-3xl`.

src/components/zen-modal/zen-modal.stories.mdx
chavlji: Then we'd need 2 separate components? `Modal` and `Modal window`?
Bug's fixed

src/components/zen-modal/zen-modal.tsx
chavlji: Then you should leave `hideCancel=false` and provide your own button in the buttons slot.

src/components/zen-modal/zen-modal.stories.mdx
chavlji: What do you mean? Each prop has a description in the props table.

src/components/zen-animate/zen-animate.tsx
chavlji: Should be async so that `transition-start` props are applied first.

src/components/zen-modal/test/zen-modals-service.spec.tsx
chavlji: Tooltips are contained in the modal itself, so the z-index of a parent modal has no impact.

src/components/zen-modal/zen-modal.stories.mdx
mplesec: Yeah, one's the component for only how the modal looks and the other would be for the functionality of modals.

src/components/zen-modal/zen-modal.stories.mdx
mplesec: Meant the instructions for how to include the modal service as well.

mplesec (approved):

PR #134, opened 2021-01-29
QUICKFIX: Add waitForElement helper function

chavlji:
Helper function waitForElement should be used inside stories to do some js after story has been rendered.
Such as appending /array props to web components. Eg. `document.querySelector('#steps').steps = args.steps`


f-jauregui (approved):
Niceeee 🔝

PR #136, opened 2021-01-29
PLAT-840: Replace zen-label and zen-input-support with zen-text

chavlji:
[JIRA ticket](https://reciprocitylabs.atlassian.net/browse/PLAT-840)


f-jauregui (approved):
Love it. Way cleaner! 🚢

PR #138, opened 2021-01-29
PLAT-963: Search bar

chavlji:
[JIRA ticket](https://reciprocitylabs.atlassian.net/browse/PLAT-963)

Screenshot 2021-01-29 at 21 14 13

Should be fixed some more, when we have icon component https://reciprocitylabs.atlassian.net/browse/PLAT-1035


f-jauregui:
@chavlji do you think it would be a good idea to merge this one before [#139](https://github.com/reciprocity/zen-ui/pull/139) and then I just do a quick PR to update icon usage?

chavlji:
@f-jauregui yeah, that was my plan. Let's merge it, and replace it later, when we'll also replace all other icons within the zen-ui.

f-jauregui (approved):
🚀

PR #145, opened 2021-02-02
PLAT-1098: ZenTable use collapsed borders

mplesec:
[JIRA ticket](https://reciprocitylabs.atlassian.net/browse/PLAT-1098)

#### Description
Managed to finally figure this out 😎

#### ChangeLOG

#### Useful links


f-jauregui:
For this specific case could be useful the zen-table-body proposal.

The only reason we see a scrollbar is because it's hardcoded for a div containing the whole table. Would be better to have a table-body component with `overflow: scroll` and maybe a height prop. In that way we would know when the content inside the table will finish.

Take a look at the next video how the scrollbar overlaps the header:

https://user-images.githubusercontent.com/48444599/106591479-17abc980-652d-11eb-9fe4-552c6b429e1d.mov

If I remove the hardcoded height like it would happen in a real use case content will leak over the header:

https://user-images.githubusercontent.com/48444599/106592476-50986e00-652e-11eb-8a63-b2c0aeafe8c1.mov

mplesec:
@f-jauregui the scrollbars overlapping the header is expected, It's part of the table. Think of it like this, if you don't set the `sticky` prop and set a fixed height, the scrollbar goes over the header in that case as well. But I get your point, it seems like a worthwhile improvement, since it would mean no external styles (the `overflow`) need to be present on a parent element. Need to talk to designers first to figure out this scrollbar thing.

f-jauregui (approved):

PR #151, opened 2021-02-05
PLAT-851: ZenBreadcrumbs component

mplesec:
[JIRA ticket](https://reciprocitylabs.atlassian.net/browse/PLAT-851)

#### Description
PR adds breadcrumbs components. Since designs are not final yet, I've left the separator as a `>` string for now and didn't bother with exact designs for links.

#### ChangeLOG

#### Useful links


f-jauregui (approved):
As we don't have clear definitions about this one I will approve. 
If we had a predefined separator (like '/') we could simplify this a bunch by doing:

```scss
::slotted(*:not(:first-child))::before {
  display: inline-block;
  padding-right: 0.5rem;
  padding-left: 0.5rem;
  color: $color-gray-200;
  content: '/';
}
```

And by doing that we can remove all this code:

```js
@Element() host: HTMLZenBreadcrumbsElement;

children: Element[];
elements: Element[];

componentWillLoad(): void {
  this.children = Array.from(this.host.children);
  this.elements = this.insertBetween<Element>(this.children, this.separatorEl());
}

insertBetween<T>(arr: Array<T>, betweenEl: T): Array<T> {
  arr = [].concat(...arr.map(el => [el, betweenEl]));
  arr.pop();
  return arr;
}

separatorEl(): Element {
  const separator = document.createElement('div');
  separator.classList.add('separator');
  separator.innerText = this.separator;
  return separator;
}
```

And after `{ reflect: true }` in the `separator` Prop, just let the render function like this:

```js
render(): HTMLElement {
  return (
    <Host>
      <slot></slot>
    </Host>
  );
}
```

Anyway, since is not defined this is the best approach and we can use whatever separator we want. Good work adding all the tests! 💯

mplesec:
@f-jauregui yeah, though it seems from the designs that for separator we might need to add an icon instead of a string, so the simpler approach could work nicely with that. And the fa icon I guess can just be put into an svg then.

mplesec:
I'm merging this since I haven't gotten a clear answer yet from designers. When changes are requested, we'll implement the simpler solution if viable.

PR #160, opened 2021-02-10
PLAT-1002: Radio button

chavlji:
[JIRA ticket](https://reciprocitylabs.atlassian.net/browse/PLAT-1002)


f-jauregui (approved):
Awesome!
Also tested in FF and Safari, great UX too.

src/components.d.ts
```diff
@@ -1027,6 +1064,32 @@ declare namespace LocalJSX {
...
+        /**
+          * Radio can't be selected (but you can still set `checked=true`)
+         */
+        "disabled"?: false;
+        /**
+          * Group id to which this radio belongs
+         */
+        "group"?: string;
+        /**
+          * Shows a red asterisk after label
```
mplesec, line 1081: The asterisk isn't red though.

src/components/zen-radio/test/zen-radio.spec.tsx
```diff
@@ -0,0 +1,158 @@
...
+    radios = Array.from(page.doc.querySelectorAll('zen-radio'));
+
+    /** Mock Input Elements focus and blur function */
+    for (const element of radios) {
+      radioInput(element).focus = jest.fn();
+      radioInput(element).blur = inputBlur;
+    }
+  });
+
+  it('renders radios', async () => {
```
mplesec, line 38: NIT: I thought we wrote these descriptions in the form of i.e. `it should render radios`

src/components/zen-radio/test/zen-radio.spec.tsx
```diff
@@ -0,0 +1,158 @@
...
+
+  it('deselects all on setting empty "selected" prop', async () => {
+    radios[2].checked = true;
+    await page.waitForChanges();
+    expect(radioInput(radios[2]).checked).toEqual(true);
+    radios[0].selected = '';
+    await page.waitForChanges();
+    for (const element of [radios[0], radios[1], radios[2]]) {
+      expect(element.checked).toEqual(false);
+    }
```
mplesec, line 68: You could use [every](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/every) for such cases.

src/components/zen-radio/test/zen-radio.spec.tsx
```diff
@@ -0,0 +1,158 @@
...
+  });
+
+  it('deselects all on setting invalid "selected" prop', async () => {
+    radios[2].checked = true;
+    await page.waitForChanges();
+    expect(radioInput(radios[2]).checked).toEqual(true);
+    radios[0].selected = 'gfdbc8';
```
mplesec, line 75: Maybe put this into a variable so it's clear what it's about.

src/components/zen-radio/zen-radio.stories.mdx
```diff
@@ -0,0 +1,109 @@
...
+        <zen-radio group="second" checked value="b">Prechecked B</zen-radio>
+      </zen-space>
+    </zen-form-group>
+  `;
+};
+
+export const StoryButtonControl = () => {
+  function changed(event) {
+    console.log('changed');
+  }
```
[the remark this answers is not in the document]
chavlji: I guess. It's nice to see when it's triggered. I've also mentioned it in the story description.

src/components/zen-radio/zen-radio.tsx
```diff
@@ -0,0 +1,155 @@
...
+  @Element() hostElement: HTMLZenRadioElement;
+
+  /** Check/uncheck radio */
+  @Prop({ reflect: true }) readonly checked: boolean = false;
+
+  /** Value of this radio option */
+  @Prop() readonly value: string = '';
+
+  /** Value of selected radio in this group */
+  @Prop() readonly selected: string = '';
```
mplesec, line 25: Should this prop be reflected as well?

mplesec (approved):

src/components/zen-radio/test/zen-radio.spec.tsx
chavlji: Not sure... I can't recall any debate on this. Currently, it's a bit of a mix, but in most cases, it's without `should`.
Let's discuss it and maybe change it in a different pr.

src/components/zen-radio/test/zen-radio.spec.tsx
chavlji: Nice! Thanks 👍🏻

src/components/zen-radio/zen-radio.tsx
chavlji: No need. `checked` is used in styling.`selected` is there just so you can get/set the result of the whole group by querying whichever element from this group. In native radio, it's called [value](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/radio#value), but since we're already using `value` prop to pass the value of the item it's impossible to also use it for that purpose, due to lack of custom prop getter in Stencil.

mplesec (approved):

PR #161, opened 2021-02-11
PLAT-1122: ZenUI Tooltip does not hide

chavlji:
[jira](https://reciprocitylabs.atlassian.net/browse/PLAT-1122)

Problem was that mouse-out happened only on the trigger element and tooltip parent element. If some child element (like some span inside tooltip), was touching the edge of the tooltip, mouse-out occurred only on this child span and not on tooltip parent (because mouse never hovered over it).
fix: mouse-out is also listened on children elements

### DIFF
--- a/src/components/zen-tooltip/zen-tooltip.tsx
+++ b/src/components/zen-tooltip/zen-tooltip.tsx
@@ -112,7 +112,8 @@ export class ZenTooltip {
 
   debounceShow = debounce(this.show, this.showDelay);
 
-  delay = this.maxHeight != 'none' ? Math.max(this.hideDelay, 150) : this.hideDelay;
+  minDelay = this.maxHeight === 'none' ? 150 : 20; // should be async!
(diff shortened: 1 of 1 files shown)


mihaelpalfi (commented):
Good work figuring it out. When testing i so that on avatar group now the hide delay is now not 0.

chavlji:
@mihaelpalfi yeah... I've changed it to min 20ms (approx 1 frame). Is it a problem?

mihaelpalfi:
> @mihaelpalfi yeah... I've changed it to min 20ms (approx 1 frame). The user shouldn't be able to notice it. Is it a problem?

Heh it just seems a little weird to me that one is still shown when i move to another one. But we had this discussion already :)

mihaelpalfi (approved):

chavlji:
@mihaelpalfi you're right... forgot about this one. I've removed the delay. It should still work.

PR #165, opened 2021-02-11
POC for popover component

mihaelpalfi:
[JIRA ticket](https://reciprocitylabs.atlassian.net/browse/)

#### Description
As per JIRA ticket

Managed to make it work. Thanks Fermin for your help. Please review.

#### ChangeLOG

#### Useful links


f-jauregui (commented):
I like the direction of this POC. Great job, Mihael! 🤙 
IMO is a good start. Maybe for the first iteration should have some event listener in case you want to close it the popover from the "target" or some button in the content inside the popover.
Would like to see what @chavlji and @mplesec think about this.

src/components/zen-popover/zen-popover.tsx
```diff
@@ -0,0 +1,79 @@
...
+
+  /** Placement */
+  @Prop() readonly offset: Offsets = { x: 0, y: 8 };
+
+  componentDidLoad() {
+    const targetSlotEl = getSlotElement(this.element, 'target');
+    const defaultSlot = getDefaultSlotContent(this.element);
+
+    if (!targetSlotEl) {
+      console.error('No target element specified for the target slot!');
```
chavlji, line 24: Imo passing target element as `prev` element is better option, because if you insist in wrapping target with your component, you ruin dom's hierarchy. I know that at least Leo would disagree with me, so we should maybe vote on this.

But in the mean time, if target slot isn't fulfilled, proclaim prev element as the target element. It's non-destructive way to support both functionalities.

src/components/zen-popover/zen-popover.tsx
```diff
@@ -0,0 +1,79 @@
...
+    }
+
+    if (!defaultSlot) {
+      console.error('No content added to default slot!');
+      return;
+    }
+
+    const defaultSlotEl = getDefaultSlotContent(this.element)[0] as HTMLElement;
+
+    const instance = createPopper(targetSlotEl, defaultSlotEl, {
```
chavlji, line 35: promote this to private class variable, because we might need it somewehere else later.

src/components/zen-popover/zen-popover.tsx
```diff
@@ -0,0 +1,79 @@
...
+        this.show(instance, defaultSlotEl);
+      } else {
+        this.hide(defaultSlotEl);
+      }
+    });
+
+    this.hide(defaultSlotEl);
+  }
+
+  show(instance, element: HTMLElement): void {
```
chavlji, line 62: when `instance` is private class var there's no need to pass it as a param

src/components/zen-popover/zen-popover.stories.mdx
```diff
@@ -0,0 +1,198 @@
...
+
+### 1. Story for a popover position
+
+- Different positioning
+
+<Canvas>
+  <Story name="StoryPositionVariants">{StoryPositionVariants.bind({})}</Story>
+</Canvas>
+
+### 2. Story for filters component
```
chavlji, line 173: Please add a story with scrollable content so we can check that popper is positioned to opposite side if there's not enough space on original popup side.

src/components/zen-popover/zen-popover.tsx
```diff
@@ -0,0 +1,79 @@
...
+    element.style.display = 'block';
```
chavlji, line 63: Add a ticket to improve animation.

src/components/zen-popover/zen-popover.tsx
```diff
@@ -0,0 +1,79 @@
...
+  shadow: true,
+})
+export class ZenPopover {
+  @Element() element: HTMLZenPopoverElement;
+
+  /** Placement */
+  @Prop() readonly placement: Placement = 'bottom-end';
+
+  /** Placement */
+  @Prop() readonly offset: Offsets = { x: 0, y: 8 };
```
chavlji, line 17: add props:
- `triggeringEvents`: [`click`, `focus`, `hover`]
- `hoverDelay`: number - timeout for how much time mouse needs to stay in one point
- `hideDelay`: number
- `visible`: `boolean | undefined` - if user passes this prop (isn't undefined) it should force show/hide popover.

All of those can be done in a separate PR.

chavlji (changes requested):
Very nice! I've added a few comments, but most of them can be changed in a separate PRs.

src/components/zen-popover/zen-popover.tsx
mihaelpalfi: Done

src/components/zen-popover/zen-popover.tsx
mihaelpalfi: Done

src/components/zen-popover/zen-popover.tsx
mihaelpalfi: Done

src/components/zen-popover/zen-popover.stories.mdx
mihaelpalfi: Added

src/components/zen-popover/zen-popover.tsx
mihaelpalfi: Added

src/components/zen-popover/zen-popover.tsx
mihaelpalfi: Added

chavlji:
> I like the direction of this POC. Great job, Mihael! 🤙
> IMO is a good start. Maybe for the first iteration should have some event listener in case you want to close it the popover from the "target" or some button in the content inside the popover.
> Would like to see what @chavlji and @mplesec think about this.

I agree, but prop `visible: boolean` would be better for this purpose instead of an event.

src/components/zen-popover/zen-popover.tsx
```diff
@@ -8,31 +8,58 @@ import { getDefaultSlotContent, getSlotElement } from '../helpers/helpers';
...
       console.error('No content added to default slot!');
       return;
     }
 
-    const defaultSlotEl = getDefaultSlotContent(this.element)[0] as HTMLElement;
+    this.defaultSlotEl = getDefaultSlotContent(this.element)[0] as HTMLElement;
+    this.createPopper();
 
-    const instance = createPopper(targetSlotEl, defaultSlotEl, {
+    window.addEventListener('click', event => {
```
chavlji, line 46: This listener should be added on popover show and removed on hide/destroy.

src/components/zen-popover/zen-popover.tsx
chavlji: Nice! It might be a good idea to make a story about that?

chavlji (changes requested):
One minor change, please.
Unit tests will be written in a different pr?

src/components/zen-popover/zen-popover.tsx
mihaelpalfi: Done

chavlji (approved):

PR #169, opened 2021-02-12
PLAT-1195: Fix number of avatar icons

mihaelpalfi:
[JIRA ticket](https://reciprocitylabs.atlassian.net/browse/PLAT-1195)

#### Description
As per JIRA ticket

image

#### ChangeLOG

#### Useful links

### DIFF
--- a/src/components/zen-avatar-group/test/zen-avatar-group.spec.tsx
+++ b/src/components/zen-avatar-group/test/zen-avatar-group.spec.tsx
@@ -74,4 +74,14 @@ describe('zen-avatar-group', () => {
     await page.waitForChanges();
     expect(page.root.shadowRoot.querySelectorAll('zen-avatar').length).toEqual(5);
   });
+
+  it('should show last two avatars in tooltip', async () => {
+    const page = await newSpecPage({
+      components: [ZenAvatarGroup, ZenTooltip, ZenAvatar, ZenAvatarIcon],
+      html: `<zen-avatar-group display-max="4"  />`,
+    });
+    page.root.users = users;
+    await page.waitForChanges();
+    expect(page.root.shadowRoot.querySelectorAll('zen-avatar').length).toEqual(4);
+  });
 });
--- a/src/components/zen-avatar-group/test/zen-avatar-group.spec.tsx
+++ b/src/components/zen-avatar-group/test/zen-avatar-group.spec.tsx
@@ -59,10 +59,7 @@ describe('zen-avatar-group', () => {
     });
     page.root.users = users;
     await page.waitForChanges();
-    expect(page.root.shadowRoot.querySelectorAll('zen-avatar').length).toEqual(4);
-    expect(
-      page.root.shadowRoot.querySelectorAll('zen-avatar')[3].shadowRoot.querySelector('zen-tooltip').childElementCount,
-    ).toEqual(2);
+    expect(page.root.shadowRoot.querySelectorAll('zen-avatar').length).toEqual(3);
   });
 
   it('should show all avatars icons', async () => {
@@ -82,6 +79,8 @@ describe('zen-avatar-group', () => {
     });
     page.root.users = users;
     await page.waitForChanges();
-    expect(page.root.shadowRoot.querySelectorAll('zen-avatar').length).toEqual(4);
+    expect(
+      page.root.shadowRoot.querySelectorAll('zen-avatar')[3].shadowRoot.querySelector('zen-tooltip').childElementCount,
+    ).toEqual(2);
   });
 });
--- a/src/components/zen-avatar-group/test/zen-avatar-group.spec.tsx
+++ b/src/components/zen-avatar-group/test/zen-avatar-group.spec.tsx
@@ -55,7 +55,7 @@ describe('zen-avatar-group', () => {
   it('should correctly render avatars', async () => {
     const page = await newSpecPage({
       components: [ZenAvatarGroup, ZenTooltip, ZenAvatar, ZenAvatarIcon],
-      html: `<zen-avatar-group display-max="3"  />`,
+      html: `<zen-avatar-group max-icons="3"  />`,
     });
     page.root.users = users;
     await page.waitForChanges();
@@ -65,7 +65,7 @@ describe('zen-avatar-group', () => {
   it('should show all avatars icons', async () => {
     const page = await newSpecPage({
       components: [ZenAvatarGroup, ZenTooltip, ZenAvatar, ZenAvatarIcon],
-      html: `<zen-avatar-group display-max="5"  />`,
+      html: `<zen-avatar-group max-icons="5"  />`,
     });
     page.root.users = users;
     await page.waitForChanges();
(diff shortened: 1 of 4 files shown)


src/components/zen-avatar-group/test/zen-avatar-group.spec.tsx
```diff
@@ -74,4 +74,14 @@ describe('zen-avatar-group', () => {
...
   });
+
+  it('should show last two avatars in tooltip', async () => {
+    const page = await newSpecPage({
+      components: [ZenAvatarGroup, ZenTooltip, ZenAvatar, ZenAvatarIcon],
+      html: `<zen-avatar-group display-max="4"  />`,
+    });
+    page.root.users = users;
+    await page.waitForChanges();
+    expect(page.root.shadowRoot.querySelectorAll('zen-avatar').length).toEqual(4);
```
mplesec, line 85: Hm, this doesn't really check it they're inside the tooltip though

src/components/zen-avatar-group/zen-avatar-group.tsx
```diff
@@ -31,12 +31,20 @@ export class ZenAvatarGroup {
...
     return this.getUserData(shownUsers);
   }
 
   hiddenUsers(): AvatarData[] {
-    const hiddenUsers = this.users.slice(this.displayMax, this.users.length);
+    let hiddenUsers = this.users.slice(this.displayMax, this.users.length);
+    // Handle edge case for displaying last two users in tooltip
+    if (this.displayMax == this.users.length - 1) {
+      hiddenUsers = this.users.slice(this.displayMax - 1, this.users.length);
+    }
```
mplesec, line 47: This won't correctly handle the case where if you have i.e. 8 users to display and the `displayMax` is either 6 or 7. In both cases it'll render 6 users and 2 in the tooltip.

mplesec (changes requested):

mplesec:
@mihaelpalfi also found a related issue (probably storybook). When changing the value of `displayMax` property the rendered items in the tooltip do not reflect the number specified in the icon.

src/components/zen-avatar-group/zen-avatar-group.tsx
mihaelpalfi: Did you test it? Becouse for me it does. This was the point of this fix. That if you have 8 users and you set show 7 icons the 7 one has +2 and 7,8 user in the tooltip.

src/components/zen-avatar-group/test/zen-avatar-group.spec.tsx
mihaelpalfi: This is the issue that i mentioned. Cant really get to the tooltip component with jest.

mihaelpalfi:
> @mihaelpalfi also found a related issue (probably storybook). When changing the value of `displayMax` property the rendered items in the tooltip do not reflect the number specified in the icon.

I can see a scroller.

image

mplesec:
> > Mihael Palfi also found a related issue (probably storybook). When changing the value of `displayMax` property the rendered items in the tooltip do not reflect the number specified in the icon.
> 
> I can see a scroller. And all 6 users in it.

Interesting, this is what happens when I try this:

https://user-images.githubusercontent.com/9392804/107772051-f7091e00-6d3b-11eb-9430-a90178ab8f53.mov

Browser: Chrome 88.0.4324.150 (x86_64)

src/components/zen-avatar-group/zen-avatar-group.tsx
mplesec: Oh, I didn't understand that. In that case, seems confusing to do that, since if the consumer specifies 7 they would expect 7 of them displayed.

src/components/zen-avatar-group/test/zen-avatar-group.spec.tsx
mplesec: Wouldn't the same approach as on L64 work?
https://github.com/reciprocity/zen-ui/blob/3a565df7253c6ef3b9717afee623cf1d438e0536/src/components/zen-avatar-group/test/zen-avatar-group.spec.tsx#L64

src/components/zen-avatar-group/zen-avatar-group.tsx
mihaelpalfi: This is a good one. Thinked about it once again and changed the implementation.  I think that QA is wrong. Based on the functional wireframes if we want to see 7 icons and you have 8. We display 6 avatars and on with +2. So together there are 7 icons displayed.

Now it should be correct.

image

src/components/zen-avatar-group/test/zen-avatar-group.spec.tsx
mihaelpalfi: Yes you are right. Changed.

mihaelpalfi:
> 
> Interesting, this is what happens when I try this:
> 
Lets try to reproduce this together.

f-jauregui:
> > > Mihael Palfi also found a related issue (probably storybook). When changing the value of `displayMax` property the rendered items in the tooltip do not reflect the number specified in the icon.
> > 
> > 
> > I can see a scroller. And all 6 users in it.
> 
> Interesting, this is what happens when I try this:
> 
>  Screen.Recording.2021-02-12.at.14.08.39.mov 
> Browser: Chrome 88.0.4324.150 (x86_64)

Experiencing same thing. Tested in Chrome, Safari and Firefox.

mihaelpalfi:
> > > > Mihael Palfi also found a related issue (probably storybook). When changing the value of `displayMax` property the rendered items in the tooltip do not reflect the number specified in the icon.
> > > 
> > > 
> > > I can see a scroller. And all 6 users in it.
> > 
> > 
> > Interesting, this is what happens when I try this:
> > Screen.Recording.2021-02-12.at.14.08.39.mov
> > Browser: Chrome 88.0.4324.150 (x86_64)
> 
> Experiencing same thing. Tested in Chrome, Safari and Firefox.

If you guys are talking about that you cant move the mouse inside the tooltip. Its again the issue because of the fix that tooltip dosent hide. I fixed the issue in another pr.

src/components/zen-avatar-group/zen-avatar-group.tsx
mplesec: I'd sooner argue we need to challenge the functional spec then. This is imo unexpected behavior for whoever will be using it.

If we do decide to keep it like this, I think we should change the prop description to reflect this. Currently it's `Max number of users to display`, which gives the wrong impression of what will happen. And maybe add a story for this behavior as well.

Also, in the description it says 5 icons is the max, but in this PR it's 8.

chavlji: The current implementation is what I'd expect. Imo it's working correctly. All misunderstandings here orignates from prop name and description. Should be:
`maxIcons` - `Max number of icons to display`

chavlji (changes requested):
Just rename the prop please and that's it.

src/components/zen-avatar-group/zen-avatar-group.tsx
mihaelpalfi: Updated param name and added description to story.

mplesec (approved):

chavlji (approved):

PR #174, opened 2021-02-15
PLAT-1139: Dropdown option - Hover style is ON when 'focused' or 'selected' are set to true even 'noHover' is set to true

chavlji:
[JIRA ticket](https://reciprocitylabs.atlassian.net/browse/PLAT-1139)


mihaelpalfi (approved):
Tested and it works as expected. Nice work.

PR #175, opened 2021-02-15
PLAT-1224: Fix tooltip padding

mihaelpalfi:
[JIRA ticket](https://reciprocitylabs.atlassian.net/browse/PLAT-1224)

#### Description
As per JIRA ticket

#### ChangeLOG

#### Useful links


src/components/zen-tooltip/zen-tooltip.stories.mdx
```diff
@@ -2,203 +2,122 @@ import { html } from 'lit-html';
...
       <div>
         <zen-button label="Info" variant="secondary"></zen-button>
         <zen-tooltip
           always-visible="true"
           position="top"
           variant="light"
-          label="This is a information tooltip."
-        ></zen-tooltip>
+        >
+            <zen-space padding="lg">
```
chavlji, line 83: Imo both `zen-space` and `zen-text` should be a part of default content... It's very tidious to have to repeat this throughout the app all the time for each small tooltip.
Just make default slot content of `zen-text` then.
For more custom cases we can have a slot named `content`. If slot `content` is fulfilled, we wouldn't render `zen-space` and `zen-text` but just this slot instead.
That would provide high standardization along with high customization when needed.

chavlji (changes requested):
Nice! It all looks good, I just have concerns regarding the convenience of such highly customizable tooltips.

src/components/zen-tooltip/zen-tooltip.stories.mdx
mihaelpalfi: Do we do this for the tooltip or in popover? Why do you think we  have to remove the tooltip?

chavlji: In tooltip. Popover's content is so specific and custom that it's expected users should provide padding. But when it comes to tooltips, 99% of tooltips focus just on the text content and should look uniformed across the apps.

What do you mean by "remove the tooltip"? I just thought we should use the popover component for the tooltip container when it's ready. This way we don't have to calculate its position manually.

mihaelpalfi: Updated that zen space wrappes default slot content. And added slot content for special contents.

chavlji (approved):

PR #176, opened 2021-02-15
PLAT-1109: FF - 'focus' and 'selected' doesn't work for dropdown

chavlji:
[JIRA ticket](https://reciprocitylabs.atlassian.net/browse/PLAT-1109)


mihaelpalfi (approved):
Tested and it works. Good work!

PR #177, opened 2021-02-15
PLAT-1223: Tooltip on scroll content hides the tooltip

chavlji:
- the problem was that mousemove doesn't occur if mouse is stall and content underneath mouse pointer is scrolled


mihaelpalfi (approved):
Works as expected!

PR #178, opened 2021-02-16
PLAT-1249: Input elements should have display inline by default

chavlji:
[JIRA ticket](https://reciprocitylabs.atlassian.net/browse/PLAT-1249)


mihaelpalfi (approved):
Checked all the stories and it looks good.

PR #180, opened 2021-02-16
PLAT-1266: Date picker 1 - Add new component with all needed elements and implement designs

chavlji:
[JIRA ticket](https://reciprocitylabs.atlassian.net/browse/PLAT-1266)


src/components/zen-space/zen-space.scss
```diff
@@ -1,12 +1,15 @@
 :host {
   box-sizing: border-box;
-  width: 100%;
   display: flex;
```
mihaelpalfi, line 3: Will this not brake something where space is used in another components and is expected to be 100%? Please check.

mihaelpalfi (approved):
For a first version it looks good! Please create additional ticket new feature and add subtasks what has to be done.

PR #181, opened 2021-02-17
PLAT-1270: Add bundle for React

mplesec:
[JIRA ticket](https://reciprocitylabs.atlassian.net/browse/PLAT-1270)

#### Description
PR adds bundle specific to React, so consumers in a React app don't have to manually wrap custom elements in React functional components. See: https://stenciljs.com/docs/react

#### ChangeLOG

#### Useful links

### DIFF
--- a/package.json
+++ b/package.json
@@ -57,6 +57,7 @@
     "@fortawesome/pro-light-svg-icons": "~5.15.1",
     "@fortawesome/pro-solid-svg-icons": "^5.15.1",
     "@stencil/eslint-plugin": "~0.3.1",
+    "@stencil/react-output-target": "0.0.9",
     "@stencil/sass": "~1.4.0",
     "@storybook/addon-actions": "~6.1.4",
     "@storybook/addon-console": "~1.2.2",
--- a/stencil.config.ts
+++ b/stencil.config.ts
@@ -1,5 +1,6 @@
 import { Config } from '@stencil/core';
 import { sass } from '@stencil/sass';
+import { reactOutputTarget } from '@stencil/react-output-target';
 
 export const config: Config = {
   namespace: 'zen-ui',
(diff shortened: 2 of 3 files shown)


leferreyra (approved):
This is pretty neat, I wonder if there's a way to export typescript types for the props

homer0:
It would be great to have this :D! I think that the only thing we should have in mind is "the custom loader" we discussed for scoping of the components.

@leferreyra "Some of the advantages of doing this are that you get types for your components" 🤘!!! (from the docs).

f-jauregui (commented):
The problem with this approach is that — according to the documentation — the output of the components transformed to React FC should be in another repository, starting from the repository cited below. The reactOutputTarget proxy should point to where this new repository is located.

> There is an example component library package available on Github so that you can get started. This repo will likely live as a sibling to your Stencil component library. https://github.com/ionic-team/stencil-ds-react-template

On the other hand, every time we make changes to zen-ui, we should also update the repository with the components transformed to React FC.

> This is the NPM package name of your core stencil package. In the case of Ionic we chose ‘@ionic/core’. This is the package that gets published that contains just your web components. This package is then used by the React package as a dependency.

If there's a way to make this work inside zen-ui, that would be great. 
I couldn't get this to work yet.

leferreyra:
@f-jauregui we can certainly help you guys maintain this, since we're doing wrappers manually anyway, I think this could be a good improvement.

f-jauregui:
> It would be great to have this :D! I think that the only thing we should have in mind is "the custom loader" we discussed for scoping of the components.
> 
> @leferreyra "Some of the advantages of doing this are that you get types for your components" 🤘!!! (from the docs).

@leferreyra  @mplesec  @homer0  🤔  what if we don't need to scope the components if we wrap it with React FC? Maybe would be the same case for Vue and Angular.

mplesec:
@leferreyra worth trying out

mplesec:
Closing this PR as it was only a quick proof of concept.

PR #191, opened 2021-02-19
PLAT-1274: Popover - Add prop visible and delays

chavlji:
[JIRA ticket](https://reciprocitylabs.atlassian.net/browse/PLAT-1274)

### DIFF
--- a/src/components/zen-popover/zen-popover.tsx
+++ b/src/components/zen-popover/zen-popover.tsx
@@ -43,9 +43,32 @@ export class ZenPopover {
 
   @Watch('visible')
   async visibleChanged(visible: boolean): Promise<void> {
+    const show = (): void => {
+      // Create popper and set display
+      this.createPopper();
+      this.popup.style.display = 'block';
+      this.visible = true;
+
+      // Add event listener for click outside
+      this.clickHandler = event => this.closeOnClickOutside(event);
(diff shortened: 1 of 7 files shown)


src/components/zen-popover/zen-popover.tsx
```diff
@@ -11,113 +11,127 @@ import { TriggerEvent } from '../helpers/types';
...
 
-  @Element() element: HTMLZenPopoverElement;
+  @Element() host: HTMLZenPopoverElement;
 
-  @State() visible = false;
+  /** Show/hide popover */
+  @Prop({ mutable: true }) visible = false;
 
   /** Placement */
   @Prop() readonly position: Placement = 'bottom-end';
```
mihaelpalfi, line 27: Change description to "Position".

mihaelpalfi (commented):
Tested everything and it works! Nice job Jan. Just a few things to do (maybe next pr?) : 
- Add unit tests.
- Add VRTs.
- Filter story alignments of components is a little of and doesn't have interactive set to true.

src/components/zen-popover/zen-popover.tsx
chavlji: Thanks! I'll do it in the following PR.

PR #192, opened 2021-02-22
PLAT-1312: Popover: Add animations

chavlji:
[JIRA ticket](https://reciprocitylabs.atlassian.net/browse/PLAT-1312)

Tests will be addressed in different tickets:
https://reciprocitylabs.atlassian.net/browse/PLAT-1313
https://reciprocitylabs.atlassian.net/browse/PLAT-1312

### DIFF
--- a/src/components.d.ts
+++ b/src/components.d.ts
@@ -342,7 +342,7 @@ export namespace Components {
          */
         "offset": Offsets;
         /**
-          * Placement
+          * Position
          */
         "position": Placement;
         /**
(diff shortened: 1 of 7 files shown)


mihaelpalfi (approved):
Works. Nice one. Can a story be added also with animation on/off, variants?

chavlji:
@mihaelpalfi currently there's no prop to disable animations. Do you think we should add one? I doubt there'll be a case where we would not want animations...

PR #193, opened 2021-02-22
PLAT-1316: ZenPanel component

mplesec:
[JIRA ticket](https://reciprocitylabs.atlassian.net/browse/PLAT-1316)

#### Description
PR adds a simple panel component. When looking up on how others did a collapse component I came across the Ant Design does it: https://ant.design/components/collapse/. And I like that approach a lot. The idea is to split the component into collapse and panel components.

Also haven't added any animations to the chevron, I guess I'll wait till that's consolidated.

#### ChangeLOG

#### Useful links


chavlji:
@mplesec I don't like this approach.

`zen-collapse` should not contain a trigger. Trigger (the thing that expands zen-collapse it) can be whatever you like (button, row, input,...). It can exist even without a trigger and can be expanded only through js.

We're already just creating a component called `zen-drawer` which is basically the same thing only animation is different (drawer slides on top, collapse pushes away existing elements).

Can you please mock this in the app for now? We'll have it ready by Wednesday.

src/components/zen-panel/zen-panel.tsx
```diff
@@ -0,0 +1,51 @@
...
+@Component({
+  tag: 'zen-panel',
+  styleUrl: 'zen-panel.scss',
+  shadow: true,
+})
+export class ZenPanel {
+  @Element() host: HTMLZenPanelElement;
+
+  /** The title of the panel */
+  @Prop() readonly header: string = 'Header';
```
chavlji, line 15: Remove this prop and add a slot `headerContent`.

chavlji, line 15: Remove this prop and add a slot `headerContent`.

src/components/zen-panel/zen-panel.tsx
```diff
@@ -0,0 +1,51 @@
...
+  /** Whether the panel has an arrow */
+  @Prop() readonly arrow: boolean = true;
+
+  @State() hidden = true;
```
chavlji, line 19: Rename it to `visible` because we have it like that in all other components.

src/components/zen-panel/zen-panel.stories.mdx
```diff
@@ -0,0 +1,83 @@
+import { Meta, Story, Canvas, ArgsTable } from '@storybook/addon-docs/blocks';
+import { html } from 'lit-html';
+
+import { getArgTypes, getDefaultArgs, getComponentData, spreadArgs } from '../../../.storybook/helpers/argTypes';
+
+import data from '../../../stencilDocs.json';
+const compData = data.components.find(n => n.tag === 'zen-panel');
+const argTypes = getArgTypes(compData);
+
+<Meta title="Containers/Panel" component="zen-panel" argTypes={argTypes} />
```
chavlji, line 10: Change to `In progress/Panel`

src/components/zen-panel/zen-panel.tsx
```diff
@@ -0,0 +1,51 @@
...
+import { faChevronRight, faChevronDown, IconDefinition } from '@fortawesome/pro-regular-svg-icons';
+
+import { applyPrefix } from '../helpers/helpers';
+
+@Component({
+  tag: 'zen-panel',
+  styleUrl: 'zen-panel.scss',
+  shadow: true,
+})
+export class ZenPanel {
```
chavlji, line 11: Why not `ZenCollapse`? Zen-panel is synonym to zen-card imo.

src/components/zen-panel/zen-panel.stories.mdx
```diff
@@ -0,0 +1,83 @@
...
+
+export const StoryWithControls = args => {
+  return html/*html*/ ` <sb-zen-panel ...="${spreadArgs(args, argTypes)}"> Content </sb-zen-panel> `;
+};
+
+export const StoryText = () => {
```
chavlji, line 16: All stories should resemble real-life designs. These stories doesn't look like design screenshot in Jira.

src/components/zen-panel/zen-panel.scss
```diff
@@ -0,0 +1,26 @@
...
+
+:host {
+  display: block;
+
+  // border-top: 1px solid $color-gray-200;
+  // margin: 0.5em 0;
+}
+
+.header-container {
+  display: flex;
```
chavlji, line 11: Add some padding to match the design.

src/components/zen-panel/zen-panel.scss
```diff
@@ -0,0 +1,26 @@
+@import '../../zen-styles/component-helpers.scss';
+
+:host {
+  display: block;
+
+  // border-top: 1px solid $color-gray-200;
```
chavlji, line 6: remove

src/components/zen-panel/zen-panel.scss
```diff
@@ -0,0 +1,26 @@
...
+  // margin: 0.5em 0;
+}
+
+.header-container {
+  display: flex;
+
+  cursor: pointer;
+}
+
+// .chevron {}
```
chavlji, line 16: remove

src/components/zen-panel/zen-panel.scss
```diff
@@ -0,0 +1,26 @@
...
+
+.header {
+  margin-left: 4px;
+}
+
+// .content {}
```
chavlji, line 22: remove

src/components/zen-panel/zen-panel.tsx
```diff
@@ -0,0 +1,51 @@
...
+  @Element() host: HTMLZenPanelElement;
+
+  /** The title of the panel */
+  @Prop() readonly header: string = 'Header';
+  /** Whether the panel has an arrow */
+  @Prop() readonly arrow: boolean = true;
```
chavlji, line 17: Do we have it anywhere in design without an arrow? If not, I'd remove this prop because I can't see a usecase.

mplesec:
@chavlji A drawer isn't the same as a collapse. A drawer slides out from an edge of a container, usually the viewport itself.

chavlji (changes requested):
On second thought, it might be useful to have such a component. We can modify it later to use the drawer component for a collapsible part.

src/components/zen-panel/zen-panel.tsx
mplesec: Updated, used just `header` for the slot name though

src/components/zen-panel/zen-panel.tsx
mplesec: Updated

src/components/zen-panel/zen-panel.stories.mdx
mplesec: Why though? Since it's a component we'll already be using in zencomply.

src/components/zen-panel/zen-panel.tsx
mplesec: My reasoning is that  a collapse component can have multiple panels (like the ant design implementation). Maybe we should ask the designers about this though.

src/components/zen-panel/zen-panel.stories.mdx
mplesec: What designs? 😅 There's none for this component here: https://www.figma.com/file/rWXEQiJcBXbkMY8V4NSM1d/ZenGRC-Component-Library

src/components/zen-panel/zen-panel.scss
mplesec: Done

src/components/zen-panel/zen-panel.scss
mplesec: Added styles for the icon

src/components/zen-panel/zen-panel.scss
mplesec: Done

src/components/zen-panel/zen-panel.stories.mdx
mihaelpalfi: Didnt you Matic post and image of the collapsible component in the ticket? I see an accordion component here https://www.figma.com/file/2qna0UZ5Qtva5vDLDKd9Gy/Access-controls-exploration-%7C-Mary---Feb?node-id=692%3A27679

mihaelpalfi (approved):
Tested and it works. Missing some animations for open and close.

src/components/zen-panel/zen-panel.stories.mdx
chavlji: Because it's incomplete. Still missing designs (we have nothing from designers yet), stories don't look ok, the component isn't polished in a sense of animations. And precisely for these reasons, it's highly likely it will be changed in the near future. If you're using it in the app, you take responsibility to maintain it in the future.

src/components/zen-panel/zen-panel.tsx
chavlji: I'll ask designers on today's meeting.

src/components/zen-panel/zen-panel.stories.mdx
chavlji: Yes, I was referring to the screenshot from the ticket. If this is the only thing we got, it should look exactly like this screenshot.

chavlji:
@mplesec shall we wait with this component for some better designs/cleaner functional specs?

mplesec:
@chavlji Yeah let's. It's not a blocker so it's fine.

src/components/zen-panel/zen-panel.stories.mdx
mihaelpalfi: What if we put it into in progress in story book and we can polish it after?

src/components/zen-panel/zen-panel.stories.mdx
```diff
@@ -0,0 +1,114 @@
...
+Expand and collapse a content area.
+
+## Component previews
+
+<br />
+
+### 1. Panel with text
+
+<Canvas>
+  <Story name="text">{StoryText.bind({})}</Story>
```
chavlji, line 78: I guess content here should be left-aligned with text in header? Talk to designers if unclear. This should be solved before using in app, because adding it later will affect already made stuff in the app.

chavlji, line 78: I guess content here should be left-aligned with text in header? Talk to designers if unclear. This should be solved before using in app, because adding it later will affect already made stuff in the app.

src/components/zen-panel/zen-panel.stories.mdx
```diff
@@ -0,0 +1,114 @@
...
+</Canvas>
+
+### 2. Panel with divs
+
+<Canvas>
+  <Story name="divs">{StoryDivs.bind({})}</Story>
```
chavlji, line 84: Not a very informative story. Remove it please.

src/components/zen-panel/zen-panel.stories.mdx
```diff
@@ -0,0 +1,114 @@
...
+</Canvas>
+
+### 3. Panel with form element
+
+<Canvas>
+  <Story name="form-element">{StoryFormElement.bind({})}</Story>
+</Canvas>
+
+### 4. Panel with form element and visible by default
```
chavlji, line 93: No need to have story for each prop. Please remove this story.

src/components/zen-panel/zen-panel.stories.mdx
```diff
@@ -0,0 +1,114 @@
...
+  <Story name="text">{StoryText.bind({})}</Story>
+</Canvas>
+
+### 2. Panel with divs
+
+<Canvas>
+  <Story name="divs">{StoryDivs.bind({})}</Story>
+</Canvas>
+
+### 3. Panel with form element
```
chavlji, line 87: No need to have story for each prop. Please remove this story.

src/components/zen-panel/zen-panel.tsx
chavlji: I still think this should be renamed to ZenCollapse. Designer were also refering to it as collapse.

chavlji (changes requested):
I guess there should also be some padding on the outside? Currently, there are none. Please ask designers.
Screenshot 2021-03-01 at 12 19 03
Paddings are crucial to be set now because we can't add them later when component is already used in the app.

src/components/zen-panel/zen-panel.stories.mdx
chavlji, line 87: Add some zen-text before and after this component so padding is clearly visible.

src/components/zen-panel/zen-panel.stories.mdx
chavlji: On second thought... if we take their design into consideration, it might be ok as it is (toggle is left aligned with arrow). If the consumer will want padding, he can wrap content into zen-space 🤷🏻

mplesec: Yeah it makes more sense to leave it to the consumer

src/components/zen-panel/zen-panel.stories.mdx
mplesec: It shows what happens with the div, i.e. what the padding is like for the content.

src/components/zen-panel/zen-panel.stories.mdx
mplesec: That's what the div story is for.

mplesec:
@chavlji will talk to designers re the naming

src/components/zen-panel/zen-panel.stories.mdx
mplesec: We got the info necessary, leaving it as is

src/components/zen-panel/zen-panel.scss
mplesec: Added

src/components/zen-panel/zen-panel.stories.mdx
mplesec: Disagree, in this case it's important to show it as a story since stories showcase core behavior.

chavlji: What core behavior? Can it be easily achieved through controls?

src/components/zen-panel/zen-panel.stories.mdx
chavlji: Is div any different than any other content? Can't this be also seen in all other stories?

chavlji:
@mplesec dont' forget to ask them about default padding of this component (my topmost comment)

src/components/zen-panel/zen-panel.stories.mdx
chavlji: I meant add zen-text before and after zen-panel so we see default outer padding of this whole component.
Screenshot 2021-03-01 at 12 58 32

src/components/zen-panel/zen-panel.stories.mdx
mplesec: The div (since it has a background color) better shows where the content boundaries end.

src/components/zen-panel/zen-panel.stories.mdx
mplesec: From https://storybook.js.org/docs/react/get-started/whats-a-story
> A story captures the rendered state of a UI component. Developers write multiple stories per component that describe all the “interesting” states a component can support.

It can also be achieved through controls, however for this component showing at first glance that it can start in this state makes sense.

mplesec:
@chavlji asked Shelley regarding these things. Her opinion is that the name should be ZenPanel. As for the padding, her answer was:
> No the padding isn't really defined other than what exists in terms of line height for the text boxes

And given that this is a similar case as with i.e. an input element, it makes sense that it's up to the consumer to define the space between form elements.

src/components/zen-panel/zen-panel.stories.mdx
mplesec: Is this within scope for these stories though?

chavlji (approved):

PR #196, opened 2021-02-23
PLAT-1317: Date picker - Open/close calendar

chavlji:
[JIRA ticket](https://reciprocitylabs.atlassian.net/browse/PLAT-1317)


mihaelpalfi (commented):
Fantastic work Jan! Just a couple of things : 

1. When changing the date in the input, the value is not set automatically in the calendar. We have to click on it. Is this intentional?

2. I managed to brake it in the way that when i clicked on the input it opens and closes  the calendar right away. Cant figure out how this happened. But i got it 3 times.

screen-capture

3. Didnt we already shorten the implementation for the popover?

chavlji:
@mihaelpalfi 
3. What do you mean?

mihaelpalfi (approved):
Tested and works as expected.

PR #198, opened 2021-02-23
PLAT-1329: Fix notification icons

mihaelpalfi:
[JIRA ticket](https://reciprocitylabs.atlassian.net/browse/PLAT-1329)

#### Description
As per JIRA ticket

@f-jauregui 
1. Should we add visible param?
2. Should we remove dismiss after timeout functionality? Is this being handled in notifications stack?

#### ChangeLOG

#### Useful links

### DIFF
--- a/src/components/zen-notification/zen-notification.tsx
+++ b/src/components/zen-notification/zen-notification.tsx
@@ -1,7 +1,8 @@
-import { Component, Host, h, Prop } from '@stencil/core';
+import { Component, Host, h, Prop, Element } from '@stencil/core';
 import { faTimes } from '@fortawesome/pro-light-svg-icons';
 import { Duration, NotificationVariant } from '../helpers/types';
 import { getIcon, getTimeout } from './helpers';
+import { applyPrefix } from '../helpers/helpers';
 
 @Component({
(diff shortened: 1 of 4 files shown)


src/components/zen-notification/zen-notification.tsx
```diff
@@ -1,15 +1,16 @@
...
 import { getIcon, getTimeout } from './helpers';
+import { applyPrefix } from '../helpers/helpers';
 
 @Component({
   tag: 'zen-notification',
   styleUrl: 'zen-notification.scss',
   shadow: true,
 })
 export class ZenNotification {
-  div: HTMLElement = undefined;
```
chavlji, line 12: I'd suggest convention to name this variable `this.host` since to be the same as in css (`:host` selector).
`element` sounds good too, so if you want we can also choose this one. It just needs to be the same in each component.

chavlji (approved):
Nice!

src/components/zen-notification/zen-notification.tsx
mihaelpalfi: In my opinion probably host would be better. But in another pr.

PR #199, opened 2021-02-23
PLAT-1269: VRT: Possible to merge PR when tests failing locally

mplesec:
[JIRA ticket](https://reciprocitylabs.atlassian.net/browse/PLAT-1269)

#### Description
PR addresses the issue where visual tests fail locally but they pass on CI and thus a PR with failing visual tests can get merged -> not ok

#### ChangeLOG

#### Useful links


cypress.json
```diff
@@ -2,8 +2,5 @@
   "projectId": "iqmzqy",
   "baseUrl": "http://localhost:6006",
   "includeShadowDom": true,
-  "video": false,
-  "retries": {
-    "runMode": 2
```
mplesec, line 7: So the problem was this line. Since it can rerun a visual test, it silent fails the 1st time but still overrides the old visual snapshot with the new inaccurate one, and then on the 2nd run effectively compares the new one with itself, which passes. Something to keep in mind when it comes to retries.

chavlji:
Great finding!

cypress.json
Donaab: @mplesec Hm ... this is odd ... why it overrides the old visual ... it should override only if you run the test with CYPRESS_updateSnapshots=true so when the tests are run with "test:e2e:update"

mihaelpalfi (approved):

PR #204, opened 2021-02-25
PLAT-1284: Date picker - Add unit tests

chavlji:
[JIRA ticket](https://reciprocitylabs.atlassian.net/browse/PLAT-1284)

### DIFF
--- a/src/components/zen-date-picker/tests/zen-date-picker.spec.tsx
+++ b/src/components/zen-date-picker/tests/zen-date-picker.spec.tsx
@@ -1,0 +1,156 @@
+import { newSpecPage, SpecPage } from '@stencil/core/testing';
+import { simulateKey } from '../../helpers/jest';
+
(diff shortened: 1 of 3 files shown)


mihaelpalfi (changes requested):
Getting an error when runing tests.

`[UnhandledPromiseRejection: This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). The promise rejected with the reason "TypeError: Cannot read property 'visible' of null".] {
  code: 'ERR_UNHANDLED_REJECTION'
}
 FAIL  src/components/zen-date-picker/tests/zen-date-picker.spec.tsx
  ● Test suite failed to run
    Call retries were exceeded
`

mihaelpalfi (approved):
Tested and date picker works as expected, test pass and good test coverage! Great work Jan.

PR #205, opened 2021-02-25
PLAT-1347: Tooltip - Rebase to popover

chavlji:
[JIRA ticket](https://reciprocitylabs.atlassian.net/browse/PLAT-1347)


mihaelpalfi (approved):
Works. Will merge this now and address some issues that i noticed in separate tickets.
 Issues : 
- We can probably remove the arrow part, dont see it on designs for no tooltip/popover and it does not work.
- Zen space alignments does not work anymore as expected (will create a ticket)
- Probably we should add a flip param so we can disable jumping of the tooltip on scroll. Now everything is jumping up and down.
- When changing params in story control it always scrolls to the middle of the page.
- Popover stories are still off. Like the filter part. Alignments ....

PR #206, opened 2021-02-26
PLAT-1359: Fixes and code cleanup

mihaelpalfi:
[JIRA ticket](https://reciprocitylabs.atlassian.net/browse/PLAT-1359)

#### Description
As per JIRA ticket

#### ChangeLOG

#### Useful links


src/components.d.ts
```diff
@@ -363,6 +363,10 @@ export namespace Components {
           * Show and hide delay. Only affects show on hover! Eg. '100' - both show & hide 100ms. '100 500' - show 100ms, hide 500ms.
          */
         "delay": string;
+        /**
+          * Popover positions it self on other side if no space available
+         */
+        "flip": boolean;
```
chavlji, line 369: As mentioned. This problem is caused by prop `alwaysVisible`. Tooltips are never always visible initially. The only use case for this is in presentational stories. It will never be used in production.
So we're adding an additional prop to fix this, that will also never be used in the production (you always want to have your tooltip shown and position doesn't matter).
ISo to fix this issue please remove prop `alwaysVisible`.

src/components.d.ts
```diff
@@ -603,10 +607,6 @@ export namespace Components {
           * Show and hide delay. Eg. '100' - both show & hide 100ms. '100 500' - show 100ms, hide 500ms.
          */
         "delay": string;
-        /**
-          * Pointing arrow - like a cartoon balloon
-         */
-        "hasArrow"?: boolean;
```
chavlji, line 609: If we need this in the future, popper supports this very nicely so it's not a problem to add.

src/components/zen-notification/test/zen-notification.spec.tsx
```diff
@@ -5,7 +5,7 @@ describe('zen-notification', () => {
   it('should set title and message', async () => {
     const page = await newSpecPage({
       components: [ZenNotification],
-      html: `<zen-notification heading="Test Heading">Test Message</zen-notification>`,
```
chavlji, line 8: 🔝

chavlji (changes requested):
Very nice! Great attention to details!
Would be great if we can agree on removing `alwaysVisible` and `flip` ;)

src/components.d.ts
mihaelpalfi: Removed always visible and flip param.

src/components.d.ts
mihaelpalfi: Agree. For now not needed.

src/components/zen-notification/test/zen-notification.spec.tsx
mihaelpalfi: Thanks

chavlji (approved):

PR #211, opened 2021-03-02
PLAT-1220: Add sortable component

mihaelpalfi:
[JIRA ticket](https://reciprocitylabs.atlassian.net/browse/PLAT-1220)

#### Description
As per JIRA ticket

#### ChangeLOG

#### Useful links


src/components/zen-draggable/zen-draggable.tsx
```diff
@@ -0,0 +1,27 @@
...
+export class ZenDraggable {
+  @Element() host: HTMLZenDraggableElement;
+
+  componentDidLoad(): void {
+    const children = Array.from(this.host.parentElement.children);
+    children.forEach(child => {
+      new Sortable(child, {
+        animation: 150,
+      });
+    });
```
mplesec, line 18: What's the reason it has to instance it for every child, instead of how it's in the [docs](https://github.com/SortableJS/Sortable#usage) where it instantiates it on the parent?

mplesec (commented):
Works fine, cool library.

Noticed one strange thing though. If you have multiple draggable components on the page the second one doesn't have any handles:
Screenshot 2021-03-03 at 11 26 28

src/components/zen-draggable/zen-draggable.tsx
mihaelpalfi: You are right. Changed.

mihaelpalfi:
> Works fine, cool library.
> 
> Noticed one strange thing though. If you have multiple draggable components on the page the second one doesn't have any handles:

Tested it and the change how sortable is initialised, fixes this issue with multiple draggables.

src/components/zen-draggable/zen-draggable.tsx
```diff
@@ -0,0 +1,22 @@
+import { Component, Host, h, Element } from '@stencil/core';
+import Sortable from 'sortablejs';
+
+@Component({
+  tag: 'zen-draggable',
+  styleUrl: 'zen-draggable.scss',
+  shadow: true,
+})
+export class ZenDraggable {
```
chavlji, line 9: Component name should be `ZenSortable`

chavlji (changes requested):
It works, but my main concern is more of a conceptual nature:
This plugin shouldn't be rearranging elements in the dom.

If you think of how reactive fe frameworks handle lists:
You have an array and when the user changes the order of the list, the array is updated. Then framework internally responds to this by re-rendering the dom representation of the list.
Your version actually moves the dom items, which shouldn't happen. Only framework should do this otherwise, you'll get collisions.

This component should:
1. take care of drag
2. draw drop indicator while dragging
3. Catch drop and send event to the app: `trigger('dropped', {oldIndex, newIndex})`

4. App will in respond to that rearrange items in the array and framework will rerender the list (probably with animations, because all fe frameworks support that ootb).

This is rather specific usecase so I doubt it is doable using sortablejs.

src/components/zen-draggable/zen-draggable.tsx
mihaelpalfi: Why? It does what the name states.

chavlji: Draggable usually refers to libs that adds support for whichever dragging (eg. dragging images across canvas, resizing stuff with drag-handles,...)
Sortable usually refers to libs that [adds support for reorderable drag-and-drop lists](https://github.com/SortableJS/Sortable#sortable------).

chavlji:
On second thought, it might be better to base it on data array instead of events:
The component takes prop `data: Any[]`
When drag happens, it changes the order of items in reference array `data`. And the fe framework will respond to this change by rerendering the dom.

But the first claim still stands: we shouldn't tamper with the dom!

mplesec:
> Tested it and the change how sortable is initialised, fixes this issue with multiple draggables.

Nice 👍 Perhaps add a story for this as well

homer0:
I don't quite understand who the implementation would be able to use it: it sends a list of nodes that can be dragged around and sorted, but there's no interface to notify that a change happened.

mihaelpalfi:
> I don't quite understand how the implementation would be able to use it: it sends a list of nodes that can be dragged around and sorted, but there's no interface to notify that a change happened.

Correct. Because it is a WIP. Also i stoped with implementation till we address Jans concerns regarding manipulating dome objects. But yeah a event must be dispatched on update. And also we have to figure out how to get a modified list of items that changed position.

homer0 (approved):
🤘 !

f-jauregui (approved):
Tested and it's all working 🚀

src/components/zen-sortable/test/zen-sortable.spec.tsx
```diff
@@ -0,0 +1,67 @@
...
+                <zen-sortable-item class="item3" data-id="3">
+                  <zen-text>Item 3</zen-text>
+                </zen-sortable-item>
+                <zen-sortable-item class="item4" data-id="4">
+                  <zen-text>Item 4</zen-text>
+                </zen-sortable-item>
+              </zen-sortable>`,
+    });
+
+    expect(Sortable.create).toHaveBeenCalledTimes(1);
```
chavlji, line 42: Please break into two tests:

`should be instantiated`
`should dispatch onChange event`

Extract things into functions, to keep it dry.

src/components/zen-sortable/zen-sortable.stories.mdx
```diff
@@ -0,0 +1,109 @@
...
+    });
+  });
+  return html/*html*/ ` <style>
+      .container {
+        width: 100%;
+      }
+    </style>
+    <sb-zen-sortable id="sortable" padding="lg" spacing="lg" style="width:230px;border:1px solid lightgray">
+      <sb-zen-sortable-item data-id="1">
+        <sb-zen-space class="container" fill horizontal-align="space-between" vertical-align="center">
```
chavlji, line 22: Remove all these `zen-space` wrapper of each item after you add `zen-space` to item component.

src/components/zen-sortable/zen-sortable.stories.mdx
```diff
@@ -0,0 +1,109 @@
...
+    </sb-zen-sortable-item>
+    <sb-zen-sortable-item data-id="4">
+      <sb-zen-text>Item 4</sb-zen-text>
+    </sb-zen-sortable-item>
+  </sb-zen-sortable>`;
+};
+
+# Zen Sortable
+
+Slotted content becomes draggable. On change position it returns an "onChange" custom event that contains list of "data-id" attributes value.
```
chavlji, line 76: Change to:
Slotted `zen-sortable-items` becomes draggable. Each draggable item should have unique attribute defined by prop `key`. On item drag&drop `drop` event is triggered. It contains list of item keys that resembles new order.

src/components/zen-sortable/zen-sortable.tsx
```diff
@@ -0,0 +1,44 @@
...
+    Sortable.create(this.host, {
+      animation: 150,
+      handle: '.handle',
+      onEnd: function (evt) {
+        evt.stopPropagation();
+        evt.target.dispatchEvent(new CustomEvent('onChange', { detail: this.toArray() }));
+      },
+    });
+  }
+  render(): HTMLElement {
```
chavlji, line 34: In a separate task create child observable and on each change check that every item has `key` attribute set. If not, raise exception.

src/components/zen-sortable/zen-sortable.tsx
```diff
@@ -0,0 +1,44 @@
...
+@Component({
+  tag: 'zen-sortable',
+  styleUrl: 'zen-sortable.scss',
+  shadow: true,
+})
+export class ZenSortable {
+  @Element() host: HTMLZenSortableElement;
+
+  /** Container padding */
+  @Prop() readonly padding: PaddingShorthand = 'none';
```
chavlji, line 19: Add property `dataIdAttr` that is just exposed sortable's option `dataIdAttr`. So user can decide which attribute he wants to use. It should have no default. Check if prop is set in `componentDidLoad` and raise expection if it wasn't (this way we force user to keep this prop in mind).

src/components/zen-sortable/zen-sortable.tsx
```diff
@@ -0,0 +1,44 @@
...
+  /** Container item spacing */
+  @Prop() readonly spacing: PaddingShorthand = 'none';
+
+  componentDidLoad(): void {
+    Sortable.create(this.host, {
+      animation: 150,
+      handle: '.handle',
+      onEnd: function (evt) {
+        evt.stopPropagation();
+        evt.target.dispatchEvent(new CustomEvent('onChange', { detail: this.toArray() }));
```
chavlji, line 30: Event names should not have `on` prepended. Since `change` is already taken and can't have payload, rename to `drop`.

src/components/zen-sortable-item/test/zen-sortable-item.spec.tsx
```diff
@@ -0,0 +1,13 @@
+import { newSpecPage } from '@stencil/core/testing';
+import { ZenSortableItem } from '../zen-sortable-item';
+
+describe('zen-sortable-item', () => {
+  it('should render', async () => {
```
chavlji, line 5: This test doesn't do much. Check that it's rendered in the shadow dom instead.

chavlji (changes requested):
Very nice! Also nice presented story!

Some minor bug:
- If I drag an item and press `Esc` dragging stops, but the item's position isn't restored. If it's not a quick fix, please create a separate ticket.

src/components/zen-sortable/zen-sortable.stories.mdx
```diff
@@ -0,0 +1,109 @@
...
+        <sb-zen-space class="container" fill horizontal-align="space-between" vertical-align="center">
+          <sb-zen-text>Manager</sb-zen-text>
+          <sb-zen-toggle></sb-zen-toggle>
+        </sb-zen-space>
+      </sb-zen-sortable-item>
+    </sb-zen-sortable>`;
+};
+
+export const StoryWithControls = args => {
+  waitForElement('#controls', element => {
```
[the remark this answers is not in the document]
mihaelpalfi: Done. and event is already documented.

src/components/zen-sortable/zen-sortable.tsx
mihaelpalfi: Created https://reciprocitylabs.atlassian.net/browse/PLAT-1584

src/components/zen-sortable/test/zen-sortable.spec.tsx
```diff
@@ -0,0 +1,67 @@
...
+  it('should render', async () => {
+    const page = await newSpecPage({
+      components: [ZenSortable],
+      html: `<zen-sortable>Content</zen-sortable>`,
+    });
+
+    expect(page.root).toBeTruthy();
+  });
+
+  it('should be instantiated', async () => {
```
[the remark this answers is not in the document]
mihaelpalfi: Leo showed me how to do it. So kudos to him.

src/components/zen-sortable-item/zen-sortable-item.tsx
```diff
@@ -0,0 +1,24 @@
...
+  styleUrl: 'zen-sortable-item.scss',
+  shadow: true,
+})
+export class ZenSortableItem {
+  @Element() host: HTMLZenSortableItemElement;
+  render(): HTMLElement {
+    const ZenIcon = applyPrefix('zen-icon', this.host);
+    return (
+      <Host>
+        <ZenIcon icon={faGripVertical} padding="md" size="sm" class="handle" />
```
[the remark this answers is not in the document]
mihaelpalfi: Display is already flex. And what is the difference wrapping it here or on zen sortable component?

src/components/zen-sortable/zen-sortable.stories.mdx
mihaelpalfi: Done. But i dont think its correct. Everything becomes slotted not only item. The uniq key is named data-id.

src/components/zen-sortable/zen-sortable.tsx
mihaelpalfi: Dont quite understand what is meant with this.

src/components/zen-sortable-item/test/zen-sortable-item.spec.tsx
mihaelpalfi: Done.

src/components/zen-sortable/zen-sortable.tsx
mihaelpalfi: Changing the event name to drop event is not picked up with logEvents?

src/components/zen-sortable/test/zen-sortable.spec.tsx
mihaelpalfi: Splited.

src/components/zen-sortable/zen-sortable.stories.mdx
homer0: - Don't use `key` as an attribute, even more so for a component that will be repeated: `key` is used internally by React on loops that generate components. Keep `data-id`.
- I agree with the part about `Slotted `zen-sortable-items` because even if you add any other element, when you instantiate `Sortable`, you give it `.handle` as the handler, so whatever they put inside should have that class.

src/components/zen-sortable/zen-sortable.tsx
homer0: @chavlji If I understand it correctly, your idea is so the implementation can add whatever they want inside the slot and with this prop control the attribute that _reflects_ the ID; but if we go that route, we should also expose a property with the class name we are using for the handle (`.handle`) and that's way too much flexibility for what we need right now.

I think the current scope (you only use `zen-sortable-item` inside `zen-sortable`) covers the case we need right now, and if we were to add this flexibility on the future, it wouldn't be that much breaking for the implementation, so we could create a TD for it.

src/components/zen-sortable/test/zen-sortable.spec.tsx
homer0: While I agree on this, we actually did it as a single test to avoid DRY: you need to do the whole setup to test the callback, and creating a function to set it up for just two methods looked like too much :P.

I'll take a look with @mihaelpalfi to see how to separate them and avoid DRY.

src/components/zen-sortable-item/zen-sortable-item.tsx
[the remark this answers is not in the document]
mihaelpalfi: Done

src/components/zen-sortable-item/zen-sortable-item.tsx
```diff
@@ -0,0 +1,24 @@
...
+/**
+ * @slot defaultSlot - Slot for custom content
+ */
+@Component({
+  tag: 'zen-sortable-item',
+  styleUrl: 'zen-sortable-item.scss',
+  shadow: true,
+})
+export class ZenSortableItem {
+  @Element() host: HTMLZenSortableItemElement;
```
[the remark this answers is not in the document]
mihaelpalfi: Done. But also added spacing.

src/components/zen-sortable/zen-sortable.stories.mdx
```diff
@@ -0,0 +1,109 @@
...
+    element.addEventListener('onChange', ev => {
+      console.log('event: ', ev);
+    });
+  });
+  return html/*html*/ ` <style>
+      .container {
+        width: 100%;
+      }
+    </style>
+    <sb-zen-sortable id="sortable" padding="lg" spacing="lg" style="width:230px;border:1px solid lightgray">
```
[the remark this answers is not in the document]
mihaelpalfi: No need. Only removed it because it defaults to none

src/components/zen-sortable/zen-sortable.stories.mdx
```diff
@@ -0,0 +1,109 @@
...
+</Canvas>
+
+## Properties
+
+<Canvas>
+  <Story
+    name="default"
+    args={{
+      ...getDefaultArgs(argTypes),
+      padding: 'lg',
```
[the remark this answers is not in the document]
mihaelpalfi: Removed

src/components/zen-sortable/zen-sortable.stories.mdx
mihaelpalfi: Removed

mihaelpalfi:
> * If I drag an item and press `Esc` dragging stops, but the item's position isn't restored. If it's not a quick fix, please create a separate ticket.

Created bug ticket https://reciprocitylabs.atlassian.net/browse/PLAT-1585
Also created the watch child ticket https://reciprocitylabs.atlassian.net/browse/PLAT-1584

src/components/zen-sortable/zen-sortable.tsx
mihaelpalfi: I agree on one point and added a check that data-id must be set for an item. If not a console error is raise.

mihaelpalfi: Should i create a ticket for "dataIdAttr" maybe?

src/components/zen-sortable/zen-sortable.stories.mdx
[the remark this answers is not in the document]
chavlji: It's still not showing up in the stories. Change `<docs-table>` in your story to `<sb-docs-table>`

src/components/zen-sortable/zen-sortable.stories.mdx
```diff
@@ -0,0 +1,87 @@
...
+    <sb-zen-sortable-item data-id="4">
+      <sb-zen-text>Item 4</sb-zen-text>
+    </sb-zen-sortable-item>
+  </sb-zen-sortable>
+  ${logEvents('#controls', getComponentData('zen-sortable'))}`;
+};
+
+# Zen Sortable
+
+Slotted `zen-sortable-items` becomes draggable. Each draggable item should have unique attribute defined by prop `key`. On item drag&drop `drop` event is triggered. It contains list of item keys that resembles new order.
```
chavlji, line 56:
```suggestion
Slotted elements becomes draggable. It's advised to use `zen-sortable-items` for slotted elements.  Each draggable item should have unique `data-id` attribute. On item drag&drop `drop` event is triggered. It contains list of `data-id`s that resembles new order.
```

chavlji: Changing it a bit, so it matches suggestions you've made in the comments. Feel free to change.

src/components/zen-sortable/zen-sortable.stories.mdx
```diff
@@ -0,0 +1,87 @@
...
+      </sb-zen-sortable-item>
+      <sb-zen-sortable-item padding="md" data-id="3">
+          <sb-zen-text>Status</sb-zen-text>
+          <sb-zen-toggle style=></sb-zen-toggle>
+      </sb-zen-sortable-item>
+      <sb-zen-sortable-item padding="md" data-id="4">
+          <sb-zen-text>Manager</sb-zen-text>
+          <sb-zen-toggle></sb-zen-toggle>
+      </sb-zen-sortable-item>
+    </sb-zen-sortable>
```
chavlji, line 28:
```suggestion
    <sb-zen-sortable id="sortable" style="width:230px;border:1px solid lightgray;">
      <sb-zen-sortable-item padding="md" data-id="1">
          <sb-zen-text>Title</sb-zen-text>
          <sb-zen-toggle class="space-before"></sb-zen-toggle>
      </sb-zen-sortable-item>
      <sb-zen-sortable-item data-id="2">
          <sb-zen-text>Description</sb-zen-text>
          <sb-zen-toggle class="space-before"></sb-zen-toggle>
      </sb-zen-sortable-item>
      <sb-zen-sortable-item data-id="3">
          <sb-zen-text>Status</sb-zen-text>
          <sb-zen-toggle class="space-before"></sb-zen-toggle>
      </sb-zen-sortable-item>
      <sb-zen-sortable-item data-id="4">
          <sb-zen-text>Manager</sb-zen-text>
          <sb-zen-toggle class="space-before"></sb-zen-toggle>
      </sb-zen-sortable-item>
    </sb-zen-sortable>
```

chavlji: Make story match the designs.

src/components/zen-sortable-item/zen-sortable-item.tsx
```diff
@@ -0,0 +1,32 @@
...
+  /** Container padding */
+  @Prop() readonly padding: PaddingShorthand = 'none';
+
+  render(): HTMLElement {
+    const ZenIcon = applyPrefix('zen-icon', this.host);
+    const ZenSpace = applyPrefix('zen-space', this.host);
+    return (
+      <Host>
+        <ZenSpace padding={this.padding}>
+          <ZenIcon icon={faGripVertical} padding="md" size="sm" class="handle" />
```
chavlji, line 26:
```suggestion
  @Prop() readonly padding: PaddingShorthand = 'sm lg';

  render(): HTMLElement {
    const ZenIcon = applyPrefix('zen-icon', this.host);
    const ZenSpace = applyPrefix('zen-space', this.host);
    return (
      <Host>
        <ZenSpace padding={this.padding}>
          <ZenIcon icon={faGripVertical} padding="md md md none" size="sm" class="handle" />
```

chavlji: Padding should default to values that match design. I realize I'm the one who said it should default to `none` in the first place, sorry...

Screenshot 2021-03-24 at 09 55 15

src/components/zen-sortable/zen-sortable.stories.mdx
```diff
@@ -0,0 +1,87 @@
...
+      ...getDefaultArgs(argTypes),
+    }}
+  >
+    {StoryWithControls.bind({})}
+  </Story>
+</Canvas>
+
+<ArgsTable story="default" />
+
+<docs-table docs={JSON.stringify(getComponentData('zen-sortable'))}></docs-table>
```
chavlji, line 87:
```suggestion
<sb-docs-table docs={JSON.stringify(getComponentData('zen-sortable'))}></sb-docs-table>
```

chavlji: docs-table is our component and needs to be prefixed. This is why it didn't show documented events.

chavlji (changes requested):
A few minor things and we can merge. I've added code suggestions. Hope it's easier this way.

I've added a [ticket for esc key](https://reciprocitylabs.atlassian.net/browse/PLAT-1614)

src/components/zen-sortable/zen-sortable.stories.mdx
mihaelpalfi: Done

src/components/zen-sortable/zen-sortable.stories.mdx
mihaelpalfi: Done

mihaelpalfi:
> A few minor things and we can merge. I've added code suggestions (as suggested by @homer0 ). Hope it's easier this way.
> 
> I've added a [ticket for esc key](https://reciprocitylabs.atlassian.net/browse/PLAT-1614)

There is already a bug ticket for this as mentioned above (https://reciprocitylabs.atlassian.net/browse/PLAT-1585) Please remove the one you created.

src/components/zen-sortable/zen-sortable.stories.mdx
mihaelpalfi: Fixed

src/components/zen-sortable-item/zen-sortable-item.tsx
mihaelpalfi: done

chavlji (approved):

PR #224, opened 2021-03-09
QUICK-FIX: Pin dependencies

mplesec:
#### Description
Due to https://github.com/reciprocity/zen-ui/pull/222, would be a good idea to avoid these kinds of problems in the future.

#### ChangeLOG

#### Useful links


chavlji (approved):

f-jauregui:
I disagree on this.
One thing is to avoid minor version changes (^), another one is to avoid all patches (~). What about security issues in the future? Shall we update update dependencies manually?
Also I don't understand why `"@babel/preset-typescript": "7.13.0"` was pinned before. Isn't this caused #222?

mplesec:
Using a bot makes more sense regarding security, like we have on ZenGRC. And we also see it in the logs.

mplesec:
Regarding the dependency, it wasn't explicitly specified in `package.json`.

chavlji:
@mplesec  But this current error with @babel/core probably appeared because we allowed minor version changes on that packages? It probably wouldn't happen if we'd allow just patch version updates?

mplesec:
I wouldn't rely on probably for this 😅 I don't think we're risking much tbh. And either a bot or some alerting can be set up in like an hour anyway.

PR #225, opened 2021-03-10
PLAT-1358: Date picker is missing a X icon to remove the date from field

chavlji:
[JIRA ticket](https://reciprocitylabs.atlassian.net/browse/PLAT-1358)


mihaelpalfi (approved):
Tested and works as expected. Nice work with the unit test!

PR #230, opened 2021-03-11
Revert "PLAT-1250: Zen space elements after first one in slot have margin left set."

chavlji:
Reverts reciprocity/zen-ui#216


mihaelpalfi (approved):
Tested. Its all back to normal.

PR #231, opened 2021-03-11
PLAT-1426: Add resize, row, cols properties

mihaelpalfi:
[JIRA ticket](https://reciprocitylabs.atlassian.net/browse/PLAT-1426)

#### Description
As per JIRA ticket

#### ChangeLOG

#### Useful links

### DIFF
--- a/src/components/helpers/types.ts
+++ b/src/components/helpers/types.ts
@@ -21,6 +21,8 @@ export type None = 'none';
 
 export type TriggerEvent = 'click' | 'hover';
 
+export type Resize = 'none' | 'both' | 'horizontal' | 'vertical' | 'initial' | 'inherit';
+
 export interface Avatar {
   userName?: string;
   email?: string;
--- a/src/components/zen-textarea/zen-textarea.tsx
+++ b/src/components/zen-textarea/zen-textarea.tsx
@@ -1,4 +1,5 @@
 import { Component, Host, h, Prop, Element } from '@stencil/core';
+import { Resize } from '../helpers/types';
 
 @Component({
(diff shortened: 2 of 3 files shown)


src/components/zen-textarea/zen-textarea.tsx
```diff
@@ -20,6 +26,9 @@ export class ZenTextarea {
   /** Placeholder of the textarea. */
   @Prop() readonly placeholder: string = null;
 
+  /** Can be resized. */
+  @Prop() readonly resizable: boolean = false;
```
chavlji, line 30: Should be `none|both|horizontal|vertical|initial|inherit` instead of boolean to match native prop exactly.

chavlji (changes requested):
Nice! One small change please to be compliant with standards...

src/components/zen-textarea/zen-textarea.tsx
mihaelpalfi: Done

src/components/helpers/types.ts
```diff
@@ -21,6 +21,8 @@ export type None = 'none';
 
 export type TriggerEvent = 'click' | 'hover';
 
+export type Resize = 'none' | 'both' | 'horizontal' | 'vertical' | 'initial' | 'inherit';
```
chavlji, line 24: nit: You could keep that type in textarea.tsx locally as we'll probably not need it anywhere else. But on the other hand, maybe it's more transparent to have all types in one place...

chavlji (approved):

PR #239, opened 2021-03-15
PLAT-1472: Remove negative margins from zen-space

chavlji:
[JIRA ticket](https://reciprocitylabs.atlassian.net/browse/PLAT-1472)

I've added space after element instead of before.
- this will fix the extra padding issue on a wrap for left-to-right languages.


mihaelpalfi (commented):
How can i reproduce this on zen space story. Tried to compare the zenui site and this pr and there was no margin set on both sides?

chavlji:
@mihaelpalfi you need to set wrap to true, spacing to lg and make row elements break into multiple rows.

mihaelpalfi (approved):

PR #240, opened 2021-03-15
PLAT-1461: Add indeterminate state

mihaelpalfi:
[JIRA ticket](https://reciprocitylabs.atlassian.net/browse/PLAT-1461)

- Added additionally indeterminate state for the tree parent rows.
- Regarding implementation a lot of room for improvement. To many loops! Will try improve it in upcoming prs.
- Test will be added here https://reciprocitylabs.atlassian.net/browse/PLAT-1537 to improve coverage.

#### Description
As per JIRA ticket

#### ChangeLOG

#### Useful links

### DIFF
--- a/src/components/zen-table-header/zen-table-header.tsx
+++ b/src/components/zen-table-header/zen-table-header.tsx
@@ -75,8 +75,9 @@ export class ZenTableHeader {
 
     this.host.parentElement.addEventListener('rowSelectChanged', () => {
       const allSelected = this.hasAllRowsSelected();
+
+      this.indeterminate = this.hasRowsSelected() && !allSelected;
       this.selected = allSelected;
-      this.indeterminate = !this.selected && this.hasRowsSelected() && !allSelected;
     });
 
     this.observer = new MutationObserver(() => this.onTableChildChanged());
--- a/src/components/zen-table-header/zen-table-header.tsx
+++ b/src/components/zen-table-header/zen-table-header.tsx
@@ -8,6 +8,7 @@ import { applyPrefix } from '../helpers/helpers';
 })
 export class ZenTableHeader {
   observer: MutationObserver = null;
+  @State() indeterminate = false;
 
   @Element() host: HTMLZenTableHeaderElement;
   @State() expandable = false;
@@ -58,6 +59,7 @@ export class ZenTableHeader {
   }
 
   onSelect(): void {
+ 	this.indeterminate = false;
     this.selected = !this.selected;
     this.headerSelectedChange.emit(this.selected);
   }
@@ -68,6 +70,14 @@ export class ZenTableHeader {
 
   componentDidLoad(): void {
     this.stickyChanged(this.sticky);
+   	this.indeterminate = this.hasSelectedRows();
+
+    this.host.parentElement.addEventListener('rowSelected', () => {
+      const allSelected = this.allSelectedRows();
+      this.indeterminate = !this.selected && this.hasSelectedRows() && !allSelected;
+      this.checked = this.selected || allSelected;
+    });
+
     this.observer = new MutationObserver(() => this.onTableChildChanged());
 
     const table = this.host.parentElement;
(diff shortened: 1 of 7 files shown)


src/components/zen-table/zen-table.tsx
```diff
@@ -1,11 +1,70 @@
...
+
+  @Listen('rowExpanded')
+  handleRowExpanded(ev: MouseEvent): void {
+    this.expandedChanged(ev.target as HTMLZenTableRowElement, Boolean(ev.detail));
+  }
+
+  @Listen('headerSelected')
+  handleHeaderSelected(ev: MouseEvent): void {
+    this.headerSelectedChanged(ev.target as HTMLZenTableHeaderElement, Boolean(ev.detail));
+  }
```
homer0, line 66: (Non blocking-)Suggestion && a question.

Suggestion: why not use the same listeners to apply the logic? Something like this:

```ts
handleRowSelected(ev: MouseEvent): void {
  const { target as HTMLZenTableRowElement, Boolean(ev.detail) } = ev;
  // ...the rest of selectedChanged(...)
}
```

Question: You use `Boolean(ev.detail)`, but if understand correctly, when the event is emitted, it already sends a boolean, is there some kind of transformation in the middle that converts them down to numbers? Because if it were a string, `Boolean('false') === true` :P

homer0 (approved):
I'm missing a lot of context, but I think I didn't see any red flags on the code :P. I tested it on local and everything seems to work as expected :D.

Did you validate whether the "sections" (first level) should implement the `indeterminate` too?

Btw, the two small comments are non-blocking.

src/components/zen-table/zen-table.tsx
mihaelpalfi: You are right on both points. Already changed the implementation in the parent pr 213. Will rebase when the on can be merged.

src/components/zen-checkbox/zen-checkbox.tsx
```diff
@@ -61,6 +61,7 @@ export class ZenCheckbox {
           disabled={this.disabled}
           checked={this.checked}
           required={this.required}
+          indeterminate={this.indeterminate}
```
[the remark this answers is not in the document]
mihaelpalfi: Hmm without setting it the indeterminate state was not set.

src/components/zen-table-header/zen-table-header.tsx
```diff
@@ -68,6 +71,14 @@ export class ZenTableHeader {
 
   componentDidLoad(): void {
     this.stickyChanged(this.sticky);
+    this.indeterminate = this.hasRowsSelected();
+
+    this.host.parentElement.addEventListener('rowSelectChanged', () => {
+      const allSelected = this.hasAllRowsSelected();
+      this.selected = allSelected;
+      this.indeterminate = !this.selected && this.hasRowsSelected() && !allSelected;
```
homer0, line 79: You are validating `!this.selected` and `!allSelected`, but on the previous line you assign `allSelected` to `this.selected`, so you are validating the same variable twice :P.

src/components/zen-table-header/zen-table-header.tsx
```diff
@@ -21,6 +21,9 @@ export class ZenTableHeader {
   /** Select all rows */
   @Prop({ mutable: true }) selected = false;
 
+  /** Checkbox indeterminate state  */
+  @Prop({ mutable: true }) indeterminate = false;
```
homer0, line 25: Is there are reason we don't _reflect_ it? Honest question :P

homer0 (approved):
Two minor comments, but seems to be working as expected 🤘

image

src/components/zen-table-header/zen-table-header.tsx
mihaelpalfi: Hmm hope i understand your question. But in this case we dont need it to be reflected in the host element. Like 
`<component indeterminate>something</component>`. Its mutable because we are setting the value of it also in the component itself.

src/components/zen-table-header/zen-table-header.tsx
mihaelpalfi: Yes you are right good eye. Removed.

homer0 (approved):
🤘

PR #241, opened 2021-03-16
PLAT-902: Implement notifications/alerts

f-jauregui:
[JIRA ticket](https://reciprocitylabs.atlassian.net/browse/PLAT-902)

#### Description
As per JIRA ticket.
We went with the global reference approach because we cannot access the referenced elements under the shadow dom of other applications when using micro frontends.
I'll assign the whole team because this kind of implementation will be used by zen-modal too. And also because I expect changes to be requested.

#### ChangeLOG

#### Useful links


src/components/zen-notification/test/helpers.spec.ts
```diff
@@ -0,0 +1,21 @@
+import { getIcon } from '../helpers';
+
+describe('zen-notification', () => {
+  it.each(['success', 'info', 'warning', 'error'])('should set title and message', async variant => {
```
homer0, line 4: I think you don't need to make it `async`, you are not returning a promise nor using `await`.

src/components/zen-notifications-wrapper/test/zen-notifications-wrapper.spec.tsx
```diff
@@ -0,0 +1,42 @@
...
+    const notification = {
+      heading: 'Title',
+      content: 'Content',
+      variant: 'success',
+      position: 'top-left',
+    };
+    page.rootInstance.displayNotification(notification);
+    expect(page.rootInstance.position).toBe('top-left');
+    expect(page.root.querySelector('zen-notification')).toBeTruthy();
+    expect(page.root.querySelector('zen-notification').getAttribute('heading')).toBe('Title');
```
homer0, line 27: If you are checking for an attribute, then you already validated that "is _truthy_".

src/components/zen-notifications-wrapper/test/zen-notifications-wrapper.spec.tsx
```diff
@@ -0,0 +1,42 @@
...
+    expect(page.root.querySelector('zen-notification').getAttribute('variant')).toBe('success');
+    expect(page.root.querySelector('zen-notification').innerText).toBe('Content');
+  });
+
+  it('should avoid rendering if notification wrapper already exists', async () => {
+    const page = await newSpecPage({
+      components: [ZenNotificationsWrapper],
+      html: `<zen-notifications-wrapper></zen-notifications-wrapper>`,
+    });
+    global.window.ZenUINotificationsWrapper = page.rootInstance;
```
homer0, line 37: If you are modifying `global`, I think we should probably add `beforeEach`/`afterEach` to "unset the change".

src/components/zen-notifications-wrapper/zen-notifications-wrapper.tsx
```diff
@@ -0,0 +1,57 @@
...
+})
+export class ZenNotificationsWrapper {
+  @Element() host: HTMLZenNotificationsWrapperElement;
+
+  /** Position of the notification */
+  @Prop({ reflect: true }) readonly position: PositionVariant = 'top-right';
+
+  /** Returns the notification wrapper reference */
+  private notificationWrapper(): HTMLZenNotificationsWrapperElement {
+    return global.window.ZenUINotificationsWrapper ? global.window.ZenUINotificationsWrapper : this.host;
```
homer0, line 24:
```suggestion
    return global.window.ZenUINotificationsWrapper || this.host;
```

src/components/zen-notifications-wrapper/zen-notifications-wrapper.tsx
```diff
@@ -0,0 +1,57 @@
+import { Component, Host, h, Prop, Element, Method } from '@stencil/core';
+import { applyPrefix } from '../helpers/helpers';
+import { PositionVariant, Notification } from '../helpers/types';
+
+declare global {
+  interface Window {
+    ZenUINotificationsWrapper: HTMLZenNotificationsWrapperElement;
```
homer0, line 7:
```suggestion
    ZenUINotificationsWrapper?: HTMLZenNotificationsWrapperElement;
```

src/components/zen-notifications-wrapper/zen-notifications-wrapper.tsx
```diff
@@ -0,0 +1,57 @@
...
+      applyPrefix('zen-notification', this.host),
+    ) as HTMLZenNotificationElement;
+
+    notification.setAttribute('heading', heading);
+    notification.setAttribute('variant', variant);
+    notification.setAttribute('dismiss', 'true');
+    notification.innerText = content;
+
+    global.window.ZenUINotificationsWrapper.setAttribute('position', position);
+    global.window.ZenUINotificationsWrapper.appendChild(notification);
```
homer0, line 42:
```suggestion
    const wrapper = this.notificationWrapper();
    wrapper.setAttribute('position', position);
    wrapper.appendChild(notification);
```

src/components/zen-notifications-wrapper/zen-notifications-wrapper.tsx
```diff
@@ -0,0 +1,57 @@
...
+  shadow: true,
+})
+export class ZenNotificationsWrapper {
+  @Element() host: HTMLZenNotificationsWrapperElement;
+
+  /** Position of the notification */
+  @Prop({ reflect: true }) readonly position: PositionVariant = 'top-right';
+
+  /** Returns the notification wrapper reference */
+  private notificationWrapper(): HTMLZenNotificationsWrapperElement {
```
homer0, line 23: Non-blocking suggestion: methods should start with a verb (like `getNotificationWrapper`), so it's easier to understand when you read the implementation.

src/components/zen-notifications-wrapper/zen-notifications-wrapper.tsx
```diff
@@ -0,0 +1,57 @@
...
+    if (this.notificationWrapper() === this.host) {
+      global.window.ZenUINotificationsWrapper = this.host;
+      return (
+        <Host>
+          <slot></slot>
+        </Host>
+      );
+    } else {
+      return null;
+    }
```
homer0, line 55:
```suggestion
    const wrapper = this.notificationWrapper();
    if (wrapper && wrapper !== this.host) return null;
    if (!wrapper) {
       global.window.ZenUINotificationsWrapper = this.host;
    }
    return (
      <Host>
        <slot></slot>
      </Host>
    );
```

- If the wrapper exists, but it's not this component, don't render.
- If the wrapper doesn't exist, assign the reference.
- If "we are here", means that "this is the wrapper" (even if the reference was set now or on the previous render), so ... render.

homer0 (commented):
The storybook looks great, but at some point we'll need to implement "expiration" && "stack limit"

src/components/zen-notifications-wrapper/test/zen-notifications-wrapper.spec.tsx
f-jauregui: Agree. I'll remove the `toBeTruthy` check, the other ones are more specific.

f-jauregui:
> The storybook looks great, but at some point we'll need to implement "expiration" && "stack limit"
Yes, didn't add that behavior to keep the PR focused. I can add it after is merged.

homer0 (approved):
itworks
 
🤘

mihaelpalfi (approved):
Works like a charm! Nice job Fermin.

src/components/zen-notifications-wrapper/zen-notifications-wrapper.tsx
```diff
@@ -0,0 +1,51 @@
...
+})
+export class ZenNotificationsWrapper {
+  @Element() host: HTMLZenNotificationsWrapperElement;
+
+  /** Position of the notification */
+  @Prop({ reflect: true }) readonly position: PositionVariant = 'top-right';
+
+  /** Displays a notification */
+  @Method()
+  async displayNotification({ heading, content, variant, position }: Notification): Promise<void> {
```
chavlji, line 24: Nice usage of named arguments

src/components/zen-notifications-wrapper/zen-notifications-wrapper.stories.mdx
```diff
@@ -0,0 +1,267 @@
...
+
+<Canvas>
+  <Story name="StoryBottomRightVariants">{StoryBottomRightVariants.bind({})}</Story>
+</Canvas>
+
+### bottom-left position
+
+<Canvas>
+  <Story name="StoryBottomLeftVariants">{StoryBottomLeftVariants.bind({})}</Story>
+</Canvas>
```
chavlji, line 267: Add `<docs-table>`. You have one prop and method that should be documented.

src/components/zen-notifications-wrapper/zen-notifications-wrapper.stories.mdx
```diff
@@ -0,0 +1,267 @@
...
+
+<br />
+
+## Component previews
+
+<br />
+
+### top-left position
+
+<Canvas>
```
chavlji, line 247: Since this component relies heavily on a method and it's usage is quite different than all the others, you should explain the basic idea in a few sentances. Also explain that we always have only one wrapper.

src/components/zen-notifications-wrapper/zen-notifications-wrapper.tsx
```diff
@@ -0,0 +1,51 @@
...
+@Component({
+  tag: 'zen-notifications-wrapper',
+  styleUrl: 'zen-notifications-wrapper.scss',
+  shadow: true,
+})
+export class ZenNotificationsWrapper {
+  @Element() host: HTMLZenNotificationsWrapperElement;
+
+  /** Position of the notification */
+  @Prop({ reflect: true }) readonly position: PositionVariant = 'top-right';
```
chavlji, line 20: Remove this prop. Since it will affect all apps no app should be allowed to change it on its own. We should reach consensus amoung designers and engineers and go with this one solution. It will look like a bug if notifications will jump around randomly...

chavlji (dismissed):
Very nice and elegant solution!

A few minor changes and it's a 🚀

src/components/zen-notifications-wrapper/zen-notifications-wrapper.tsx
homer0: I would consider this a non-blocker issue; but I partially agree with what @chavlji said, so I asked Shelly if she could validate it.

PR #245, opened 2021-03-17
PLAT-1491: Reflect the change of the progress tracker active index

homer0:
[PLAT-1491: Progress tracker not sending information on the "change" event - JIRA](https://reciprocitylabs.atlassian.net/browse/PLAT-1491)

Hey! This is a draft because I didn't remove the "internal state" (`internalActiveIndex`) because it seems to be tied to the "items state", and when removed, only 2 out of 3 steps would trigger the event (:P)... I'm sure I was missing some context there, so I went with just mutating the prop and leave the method that watches the prop update the internal one.

By mutating the value, I didn't need to add a custom event, as the implementation can check the `target.activeIndex`.


f-jauregui (approved):
🚀

f-jauregui (approved):
Now for real ✅

PR #247, opened 2021-03-18
PLAT-1199: Fix Husky, lint-staged, ESLint && Prettier

homer0:
[JIRA ticket](https://reciprocitylabs.atlassian.net/browse/PLAT-1199)

#### Description

- I updated some of the tooling dependencies.
- I fixed `lint-staged` implementation: It was not only disabled, but if enabled, it would lint THE ENTIRE PROJECT :P.
- I enabled `commitlint` and moved to the `commit-msg` hook.
- Downside, you can't run `yarn lint` or `yarn format` without specifying the path of what you want to lint/fix... but having them on the pre-commit (and on the CI), that becomes an edge case.
- I updated both the `README` and the contribution guidelines to reflect the changes.
- No, I didn't remove `prettier/@typescript-eslint`, they simplified the way it works on their new version, so it's no longer necessary to add it.

I wanted to add type validation on the pre push too (like we did on reciprocity/zencomply#150), but there were too many typing errors, so I'll leave that for later.


f-jauregui (approved):
Awesome work homer0! 🎩

mihaelpalfi (approved):
LGTM

PR #248, opened 2021-03-21
PLAT-1541: Add  properties watch

mihaelpalfi:
[JIRA ticket](https://reciprocitylabs.atlassian.net/browse/PLAT-1541)

#### Description
As per JIRA ticket

- Moved row selected logic from table to row watch selected
- Moved row expanded logic from table to row watch selected
- Moved table header select logic to header watch selected
- Cleaned up the components code and added descriptions

#### ChangeLOG

#### Useful links

### DIFF
--- a/src/components/zen-table-header/zen-table-header.tsx
+++ b/src/components/zen-table-header/zen-table-header.tsx
@@ -26,7 +26,7 @@ export class ZenTableHeader {
   @Prop({ mutable: true }) selected = false;
 
   /** Checkbox indeterminate state  */
-  @Prop({ mutable: true }) indeterminate = false;
+  @Prop({ mutable: true }) $indeterminate = false;
 
   @Watch('selected')
   async selectedChanged(selected: boolean): Promise<void> {
@@ -59,7 +59,7 @@ export class ZenTableHeader {
   }
 
   hasAllRowsSelected(): boolean {
-    return this.allRows().every(row => row.selected && !row.indeterminate);
+    return this.allRows().every(row => row.selected && !row.$indeterminate);
   }
 
   onSelect(): void {
@@ -78,7 +78,7 @@ export class ZenTableHeader {
 
   componentDidLoad(): void {
     this.toggleStickyChildren(this.sticky);
-    this.indeterminate = this.hasRowsSelected();
+    this.$indeterminate = this.hasRowsSelected();
 
     this.host.parentElement.addEventListener('rowSelectChanged', () => {
       const allSelected = this.hasAllRowsSelected();
@@ -89,7 +89,7 @@ export class ZenTableHeader {
       } else if (allSelected) {
         this.selected = true;
       }
-      this.indeterminate = someSelected && !allSelected;
+      this.$indeterminate = someSelected && !allSelected;
     });
 
     this.observer = new MutationObserver(() => this.onTableChildChanged());
@@ -113,7 +113,7 @@ export class ZenTableHeader {
           <div class="widgets">
             <ZenCheckBox
               class="checkbox"
-              indeterminate={this.indeterminate}
+              indeterminate={this.$indeterminate}
               checked={this.selected}
               onClick={() => this.onSelect()}
             ></ZenCheckBox>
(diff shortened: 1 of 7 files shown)


src/components/zen-table-header/zen-table-header.tsx
```diff
@@ -49,35 +57,40 @@ export class ZenTableHeader {
...
+      const someSelected = this.hasRowsSelected();
 
-      this.indeterminate = this.hasRowsSelected() && !allSelected;
-      this.selected = allSelected;
+      if (!someSelected) {
+        this.selected = false;
+      }
+      if (allSelected) {
+        this.selected = true;
+      }
```
homer0, line 92: I meesed up... I was reviewing only one commit :P.

I already left two comments regarding this kind of `if`s, it should be a couple of files below.

homer0 (approved):
Seems to keep working as expected 🤘 - I just left a single comment (multiple times) regarding some validations :P.

src/components/zen-table-row/zen-table-row.tsx
```diff
@@ -2,6 +2,13 @@ import { h, Component, Host, Prop, Element, Event, EventEmitter, Method, Watch }
 import { applyPrefix } from '../helpers/helpers';
 import { faChevronRight } from '@fortawesome/pro-light-svg-icons';
 
+/**
+ * @slot [default] - Content for table cells
+ * @event rowSelectChanged | Called on any selection change
+ * @method hasRowsSelected | Returns true if descendents have any row selected
+ * @method hasAllRowsSelected | Returns true if descendents have all rows selected
+ * @method parentRow | Returns parent row
```
chavlji, line 10: You don't have to manually document methods and events created with Stencil's decorators `@Method` and `@Event`. Just slots and events triggered manually with native dispatch.
```suggestion
```

chavlji (changes requested):
Very nice! Works as expected!

I've spotted little inconsistecy:
Sometimes clicking indeterminate checkbox checks all children, while other time it unchecks all of them. It should always select all when clicking indeterminate checkbox.

https://user-images.githubusercontent.com/5729421/112620563-009ca000-8e29-11eb-9265-5a56bf6f5484.mov

Another question: Do we still need prop `row.indeterminate`? If I set it programmatically it just messes the state. I guess that's again the question of who's controlling the state. But in this case I think the component is in charge, so there's no point to expose it to the app.

homer0:
To @chavlji 's comment:

> Another NIT: Do we still need prop row.indeterminate? If I set it programmatically it just messes the state. I guess that's again the question of who's controlling the state. But in this case I think the component is in charge, so there's no point to expose it to the app.

If removing it simplifies the state, go for it and we can think of `indeterminate` as a computed value :D.

src/components/zen-table-row/zen-table-row.tsx
mihaelpalfi: Removed

homer0:
- The table row is still showing the `indeterminate` prop on the story.
- Not sure if the tests are failing because of the change or if it's a random error :P.

mihaelpalfi:
> * The table row is still showing the `indeterminate` prop on the story.
> * Not sure if the tests are failing because of the change or if it's a random error :P.

1. Removed the prop from row story.
2. Fixed header story so test are now ok.

chavlji (approved):
I'm approving this, just for sake to get it over and merge it finally.

But I really strongly disagree with hiding props. Consumer has now no way to find out that he can at least read props `expended`, `depth`, `visible`,... they're not not documented anywhere. Sole purpose of storybook is to automate these docs and we're now adding a bunch of expectations that adds to confusion. If we don't want enduser to set them, prefix them with `$` or mention in the description that it's preferable to not set those props.

I've added tasks to improve these things:
https://reciprocitylabs.atlassian.net/browse/PLAT-1667
https://reciprocitylabs.atlassian.net/browse/PLAT-1668

mihaelpalfi:
> I've added tasks to improve these things:
> https://reciprocitylabs.atlassian.net/browse/PLAT-1667
> https://reciprocitylabs.atlassian.net/browse/PLAT-1668

Good Jan. now its in my opinion a little more manageable, so that we dont have everything in one pr.

chavlji:
@mihaelpalfi I agree!

PR #250, opened 2021-03-22
PLAT-1453: REQUEST: Datepicker default

chavlji:
[JIRA ticket](https://reciprocitylabs.atlassian.net/browse/PLAT-1453)

- added watch for `formatted-date`
- added story for this and all tests


src/components/zen-date-picker/zen-date-picker.stories.mdx
```diff
@@ -2,34 +2,68 @@ import { Meta, Story, Canvas, ArgsTable } from '@storybook/addon-docs/blocks';
...
+  });
   return html/*html*/ `
     <sb-zen-date-picker id="date-1" style="width: 192px" ...="${spreadArgs(args, argTypes)}"> </sb-zen-date-picker>
     ${logEvents('#date-1', getComponentData('zen-date-picker'))}
   `;
 };
 
+export const initialValue = () => {
+  return html/*html*/ `
+    <sb-zen-space spacing="lg">
```
mihaelpalfi, line 29: Spacing between elements does not work anymore? Something for a different PR.

src/components/zen-date-picker/zen-date-picker.tsx
```diff
@@ -196,7 +213,11 @@ export class ZenDatePicker {
   }
 
   componentDidLoad(): void {
-    this.dateChanged(this.value);
+    if (this.formattedDate !== 'null' && this.formattedDate !== null) {
```
mihaelpalfi, line 216: Why is this " !='null' " needed? It should not be possible to set null as a string. Is it because Story book?

src/components/zen-date-picker/zen-date-picker.tsx
```diff
@@ -69,6 +70,22 @@ export class ZenDatePicker {
...
 
+  @Watch('formattedDate')
+  async formattedDateChanged(formatted: string): Promise<void> {
+    const parsedDate = parseDate(formatted, this.format);
+    if (isSameDay(parsedDate, this.value)) return;
+
+    if (isValid(parsedDate)) {
+      this.value = parsedDate;
+    } else if (this.allowEmpty) {
+      this.clearDate();
```
mihaelpalfi, line 81: Should this not have await?

src/components/zen-date-picker/zen-date-picker.tsx
```diff
@@ -196,7 +213,11 @@ export class ZenDatePicker {
   }
 
   componentDidLoad(): void {
-    this.dateChanged(this.value);
+    if (this.formattedDate !== 'null' && this.formattedDate !== null) {
+      this.formattedDateChanged(this.formattedDate);
+    } else {
+      this.dateChanged(this.value); // set today date
```
mihaelpalfi, line 219: Should this not have await?

mihaelpalfi (changes requested):
Works. Just a couple of small comments.

src/components/zen-date-picker/zen-date-picker.stories.mdx
chavlji: Yeah... this pr will fix it: https://reciprocitylabs.atlassian.net/browse/PLAT-1494

src/components/zen-date-picker/zen-date-picker.tsx
chavlji: Yup. Jest or Storybook set it to 'null' instead null. Not sure which one...

src/components/zen-date-picker/zen-date-picker.tsx
chavlji: Since it's right before the return I guess there's no need to wait?

src/components/zen-date-picker/zen-date-picker.tsx
chavlji: No, because we're not relaying on it's result.
Another thing to mention is that this function is actually synchronous... it's marked as `async` only because all Stencil's `@methods` need to be marked as async.

mihaelpalfi (approved):

PR #253, opened 2021-03-23
PLAT-1538: REQUEST: Control the padding on the modal slots

chavlji:
[JIRA ticket](https://reciprocitylabs.atlassian.net/browse/PLAT-1538)


homer0:
I understand where are you trying to go, but I don't think this is the best solution. If the implementation can't control the padding, the default should be _none_: based on the use cases we have right now, zencomply will use the `paddingless-` slots, but if they use them, they don't get the footer button nor the close icon.

Another thing that could do is to create the attributes `header-padding`, `content-padding` and `footer-padding` and forward them to the `zen-space` components that wrap each slot. If I'm not wrong, the only requirement right now is with the header, so we'll kind of cover our default use case and the customization needed by the implementation.

We can talk about this tomorrow :D.

Btw, a few small extra SUPER-NON-BLOCKING details:

- The icon for the modal should be inside a button or have the `role="button"`.
- Use `feat` not `feature` for the commits.
- The scope for the commits, like on `feature: zen-space - size xl to 24 px` should be wrapped in parentheses: `feat(zen-space): size xl to 24 px`.
- You are mixing `test` and `tests` on the commits.

chavlji:
You're right. Not getting default close or ok buttons is an issue. Setting default padding to none here isn't ok I guess, because we have exact specs from designers. All components should look as per designer's specs ootb.

I like the approach with exposing extra padding props for each part! It also follows the same philosophy as all our other components (they'll support similar `padding` prop). I'll add those props to modal now.

As for commit messages, noted! Didn't know the thing of how the scope should be defined. Thanks! I'll check some conventional commits manual now.

mihaelpalfi (commented):
Looks nice but code is a little complicated with all this slots. Why not just wrap header, footer, content in zen-space and add props so that we can set the paddings. Default none.

And also remembered that we also have to change the displaying of the modal like we did for notifications. Right? If yes and there is no ticket then we should create it.

chavlji:
I agree it's too complicated. I'll change it as you suggest (expose padding of zen-space as props).

@mihaelpalfi what do you mean with "have to change the displaying of the modal like we did for notifications"? Imo the way modal currently works is ok?

mihaelpalfi (commented):
Good work. Found this issue but probably not related to this issue. 
image
Padding to large

src/stories/changelog.stories.mdx
```diff
@@ -7,6 +7,10 @@
 #### Breaking changes
 
 1. Prop `padding` value `xl` was changed from `2rem` to `1.5rem`. New prop value `2xl` was added that equals `2rem`. So to keep the same padding, replace `padding="xl"` with `padding="2xl"`
+1. `zen-modal`: For consistency reasons, the following slots were renamed:
+
+- `footer` to `paddingless-footer`
+- `buttons` to `footer`
```
homer0, line 13: Remember to update this.

(Btw, why did you use two `1.` :P? I've been checking docs and all the lists only use `1.` for items xD)

src/components/zen-modal/zen-modal.scss
```diff
@@ -48,29 +46,22 @@ $show-time: 0.3s;
...
     opacity: 0;
     transform: translate(0, -30px);
   }
 }
 
 .header {
-  display: flex;
-  align-items: center;
-  padding: 1rem 0 1rem $padding-x;
+  border-bottom: 1px solid $color-gray-200;
```
homer0, line 57: Not sure in which commit this was added, but it should have something like this:

```js
BREAKING CHANGE: The modal header now has a border bottom
```

... or similar; that's a fix from our point of view, but a breaking change for the implementation.

homer0 (approved):
🤘!

mihaelpalfi (approved):
Great animations Jan. Tested and setting the paddings works. One small NIT : 
The close button gets pushed down on long title names.

image

src/stories/changelog.stories.mdx
chavlji: md takes care of this and converts it to correct numbers. No need to keep tracks of that manually:

src/components/zen-modal/zen-modal.scss
chavlji: It's not a breaking change. It's just an improvement by designers.

chavlji: Agree with that yeah. Noted!

PR #255, opened 2021-03-23
PLAT-1593: Unify padding prop and add quick props

chavlji:
[JIRA ticket](https://reciprocitylabs.atlassian.net/browse/PLAT-1593)
It's subtask of this [main task](https://reciprocitylabs.atlassian.net/browse/PLAT-1464)
Also check commit messages for details.


.storybook/helpers/argTypes.js
```diff
@@ -23,6 +23,20 @@ function getDefaultValue(prop) {
...
+  const paddingProps = ['p', 'px', 'py', 'pt', 'pr', 'pb', 'pl'];
+
+  Object.keys(argTypes).forEach(prop => {
+    if (!paddingProps.includes(prop)) return;
+    if (prop === 'p') {
+      argTypes[prop].description = descript;
+    } else {
+      delete argTypes[prop];
+    }
+  });
```
homer0, line 37: Not a blocker; it's always recommended to avoid mutation by reference and mutation on a `forEach` (when you have the other array methods):

```suggestion
  return Object.keys(argTypes).reduce((acc, prop) => {
    if (!paddingProps.includes(prop)) return acc;
    if (prop === 'p') {
      acc[prop].description = descript;
    } else {
      delete acc[prop];
    }
  }, { ...argTypes });
```

1. Instead of mutating the reference, it creates a shallow copy to work with.
2. It deletest on the accumulator directly.

src/components/helpers/helpers.scss
```diff
@@ -0,0 +1,56 @@
+// Padding --------------------------------------
+
+$spacings: (
+  'none': 0,
+  'xs': 1px,      // 1 px
```
homer0, line 5: Question: is there a reason this is not `0.16rem`?

homer0 (commented):
Looks good, but I have a few comments/questions:

1. Does a few bytes justifies changing `padding` to `p`? I see the point of reducing the bundle size, but if does it make sense to sacrifice a good "interface" for just a few bytes? Added to the overhead that will mean when implementing a component: known which props are a single letter and which aren't. I'm ok if everyone is ok with it.
2. It does say that it supports the quick forms like `px`, `py`, `pl`, etc.... but I didn't find where does it says what they mean :P.
3. You didn't add the `BREAKING CHANGE:` comment on the breaking commits. That's ok for now, as we haven't implemented `semantic-release` yet, but it's very important to have it in mind, as it will affect the version bump and the changelog.

src/components/helpers/helpers.scss
chavlji: Maybe it's not so consistent, but 1 px is sometimes useful, while 2.56 px isn't?

.storybook/helpers/argTypes.js
chavlji: Agree. Though above code isn't working as expected:
Screenshot 2021-03-24 at 09 00 34
I've added a new commit where I just iterate through shallow copy's keys.

chavlji:
@homer0 
1. Maybe you're right...I wanted to model [Bootstrap spacing helpers](https://getbootstrap.com/docs/4.5/utilities/spacing/). It's not about reducing bundle size so much as to keep consumer markup more readable. So this:
`<zen-space vertical spacing="sm" px="lg" pt="sm" pb="xl">`
instead of
`<zen-space vertical spacing="sm" padding-horizontal="lg" padding-top="sm" padding-bottom="xl">`
I agree it's a bit confusing for the consumer in the beginning, but since this feature will be used a lot, I think it's worth it?
For now, I'm keeping a short version. I'll open a poll on the`#zen-ui` channel to reach a consensus.
2. Added explanation
3. Noted!

src/components/helpers/helpers.scss
homer0: I know this code is outdated now, but `0.16rem` is `1px`, as the browser defaults is `16px` (`1rem`)... unless we are changing it, but can it be changed from code? I thought that was only possible with `em`.

.storybook/helpers/argTypes.js
homer0: Is it possible that `argTypes` is `undefined` on some point? because that error looks like it's being generated by `Object.keys` or the spread on the accumulator.

homer0 (approved):

PR #256, opened 2021-03-23
QUICKFIX: re-enable lint-staged

homer0:
No Jira - Follow-up fix for #254 - We were trying to format EVERYTHING with ESLint, so I changed the config of `lint-staged` so non-t/js files are only parsed by Prettier.

@chavlji The reason #255 is failing it's because one of the files is missing Prettier formatting. You can run `yarn format src/` to fix it.


chavlji (approved):
That's great! 😎

PR #258, opened 2021-03-24
PLAT-1601: Date picker is not closed properly if user type the date

chavlji:
- picker was reopened because of input.focus. Focusing input after click isn't needed anyway.

[JIRA ticket](https://reciprocitylabs.atlassian.net/browse/PLAT-1601)


mihaelpalfi (approved):
Works

PR #260, opened 2021-03-24
PLAT-1371: Panel - Add animations

chavlji:
[JIRA ticket](https://reciprocitylabs.atlassian.net/browse/PLAT-1371)


src/components/zen-panel/test/zen-panel.spec.tsx
```diff
@@ -1,5 +1,11 @@
 import { newSpecPage } from '@stencil/core/testing';
 
+import * as helpers from '../../helpers/animations';
+helpers.showInstantly = jest.fn();
+helpers.showWithAnimation = jest.fn();
+helpers.hideWithAnimation = jest.fn();
+helpers.hideInstantly = jest.fn();
```
homer0, line 7: 1. Do you need any function from `animations` not to be mocked? If not, you can just `jest.mock('../../helpers/animations') on top of the file and Jest will take care of creating the mocks.
2. It's always best to reset/clear mocks before each test, so you can properly _test_ that they are called the number of times you expect and they don't affect other tests (a mock that is not resetted/cleared will keep the same state between tests).

src/components/zen-panel/test/zen-panel.spec.tsx
```diff
@@ -11,50 +17,51 @@ describe('zen-panel', () => {
...
+    const render = async (attributes: string) => {
+      page = await newSpecPage({
+        components: [ZenPanel],
+        html: /*html*/ `
+        <zen-panel ${attributes} />
+      `,
+      });
+      await page.waitForChanges();
+      panel = page.root as HTMLZenPanelElement;
+    };
```
homer0, line 33: You should put this on a `beforeEach` if possible, so no test can cause side effects on another.

homer0 (approved):
Works great 🤘!!

- The comments about the tests are not blockers ;)
- There are two SUPER tiny details that would be great if you could take a look, otherwise, I think we are good here:
  - When I toggle the panel, the cursor changes to the one for selecting text until I move it and then go back to pointer.
  - This is something I think I can only notice because of the boxes of Storybook on top of the dark background: since we are using `visibility` for this, when you toggle the panel, you can see the container changing its size before the animation starts. It's most noticeable when closing the panels. Maybe we can do something like `height: 0` to `height: auto` (I know it involves a workaround/hack :P) to make it look more _"smooth"_.

Again, none of my comments are blockers, this is already great as it is :D.

src/components/zen-panel/test/zen-panel.spec.tsx
```diff
@@ -11,50 +17,51 @@ describe('zen-panel', () => {
...
-      expect(panel.internalVisible).toBe(false);
-      expect(panel.icon().iconName).toEqual('chevron-right');
-    });
-
-    it('should display down chevron icon if visible', () => {
-      const panel = new ZenPanel();
-
-      panel.toggleContent();
+  describe('content', () => {
+    let page: SpecPage;
```
mihaelpalfi, line 21: SpecPage not imported.

mihaelpalfi (approved):
Works like a charm. Tried on all browser. Good work. Just found i minor thing in tests.

src/components/zen-panel/test/zen-panel.spec.tsx
chavlji: Done. Thanks for the tips!

src/components/zen-panel/test/zen-panel.spec.tsx
chavlji: I can't because render() takes one argument, that I can't pass to beforeEach function?

src/components/zen-panel/test/zen-panel.spec.tsx
chavlji: Fixed.

chavlji:
@homer0
- changing cursor: I can't reproduce that. Where does that happen? On a button?
- It's very tricky to animate height... first, it's not hardware accelerated, second, if you set `height:auto`, transitions don't work (you need to specify height in pixels or do hacks like animating max-height and setting it to some enormous number)

PR #262, opened 2021-03-25
PLAT-1586: Fix table sticky background

mihaelpalfi:
[JIRA ticket](https://reciprocitylabs.atlassian.net/browse/PLAT-1586)

- fixes the issue that rows z-index header gets messed up and table header sticky is not shown above the rows. (basicly the transform: scale(1) hack)
image

- refactored the sticky part so that the param is reflective now and class is applied also to children.

#### Description
As per JIRA ticket

#### ChangeLOG

#### Useful links

### DIFF
--- a/src/components.d.ts
+++ b/src/components.d.ts
@@ -504,12 +504,6 @@ export namespace Components {
          */
         "value": string;
     }
-    interface ZenSkeleton {
-        /**
-          * Makes the skeleton rounded
-         */
-        "rounded": boolean;
-    }
     interface ZenSortable {
         /**
           * Container padding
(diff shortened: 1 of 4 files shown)


chavlji:
We should avoid using z-indexes on everything except popovers and modals.

Here's an example of such problems with this version:
Row that holds dropdown has z-index 2. Header also has z-index 2. Because header is listed after row in the dom tree, it is rendered above row. And since dropdown is on same layer as row (row is promoted to layer because of z-index: 2), it is covered by the header:

Screenshot 2021-03-25 at 16 30 58

There are only 2 solutions to this problem:
1. Attach modals, popovers,... directly to body (but that might mess reactive frameworks)
2. Don't use z-indexes except for modals, popovers,...

chavlji:
Another example of this problem:
If ordinary row has dropdown, both detailed row and header covers dropdown.
Screenshot 2021-03-25 at 16 42 36

mihaelpalfi:
> Another example of this problem:
> If ordinary row has dropdown, both detailed row and header covers dropdown.
> Screenshot 2021-03-25 at 16 42 36

This second one is a problem yes. Will check further.

chavlji (approved):
👍

PR #263, opened 2021-03-25
QUICK: Add lint:all and format:all commands

chavlji:
Add commands for fixing format and lint errors.


mihaelpalfi (approved):
Good. Wanted to add this as-well.

PR #264, opened 2021-03-25
PLAT-1606: Add avatar details variant basic

mihaelpalfi:
[JIRA ticket](https://reciprocitylabs.atlassian.net/browse/PLAT-1606)

#### Description
As per JIRA ticket

image

#### ChangeLOG

#### Useful links


src/components/zen-avatar-details/zen-avatar-details.tsx
```diff
@@ -28,26 +34,53 @@ export class ZenAvatarDetails {
   /** Spacing between icon and username  */
   @Prop() readonly spacing: Size = 'md';
 
+  textSize(): string {
+    let textSize = 'md';
+    if (this.variant == 'detailed') return textSize;
```
homer0, line 39:
```suggestion
    if (this.variant === 'detailed') return textSize;
```

src/components/zen-avatar-details/zen-avatar-details.tsx
```diff
@@ -28,26 +34,53 @@ export class ZenAvatarDetails {
   /** Spacing between icon and username  */
   @Prop() readonly spacing: Size = 'md';
 
+  textSize(): string {
```
homer0, line 37: This is just a suggestion that I consider good practice: use verbs as prefixes for methods/functions (like get/set/change/update/etc.) and leave nouns for properties && getters/setters. For example, in this case, it would be `getTextSize`.

src/components/zen-avatar-details/zen-avatar-details.tsx
```diff
@@ -28,26 +34,53 @@ export class ZenAvatarDetails {
...
+      case 'sm':
+        textSize = 'md';
+        break;
+      case 'md':
+        textSize = 'lg';
+        break;
+      case 'lg':
+        textSize = 'xl';
+        break;
+    }
```
homer0, line 51: You could implement a private _map_ for these, something like this:

```js
const basicTextSizeByDetailSize = {
  sm: 'md',
  md: 'lg',
  lg: 'xl',
};
```

And then just return `basicTextSizeByDetailSize[this.variant]`.

src/components/zen-avatar-details/zen-avatar-details.tsx
```diff
@@ -28,26 +34,53 @@ export class ZenAvatarDetails {
...
     const ZenAvatarIcon = applyPrefix('zen-avatar-icon', this.host);
     const ZenSpace = applyPrefix('zen-space', this.host);
     const ZenText = applyPrefix('zen-text', this.host);
     return (
       <Host>
-        <ZenSpace no-wrap padding={this.padding} vertical-align="middle" spacing={this.spacing}>
+        <ZenSpace
+          no-wrap
+          padding={this.padding}
+          vertical-align={this.variant == 'basic' ? 'center' : 'start'}
```
homer0, line 65: I'm seeing too many ternaries; you can do something like the sizes dictionary I suggested on the method for the text size:

```js
const propsByVariant = {
  basic: {
    mainVerticalAlignment: 'center',
    avatarSize: null, // so you can short circuit with this.size (prop.avatarSize || this.size),
    usernameBold: false,
  },
  detailed: {
    mainVerticalAlignment: 'start',
    avatarSize: 'md',
    usernameBold: true,
  },
}
```

It will not only get rid of the ternaries, but it will reduce the number of possible branches and reaching good coverage will be easier 😉 .

homer0 (commented):
It looks good overall 🤘, just left a few small comments on the "execution".

src/components/zen-avatar-details/zen-avatar-details.tsx
mihaelpalfi: Heh this is my opinion as well. But on other prs Jan dosent want this naming. Maybe my coding style is to old school and this has changed in the years. I will change it.

src/components/zen-avatar-details/zen-avatar-details.tsx
```diff
@@ -28,26 +34,53 @@ export class ZenAvatarDetails {
...
+          spacing={this.spacing}
```
chavlji, line 66: I realy strongly believe we should remove prop `spacing`. In all official designs it seems like variant basic has spacing 8 px and detailed variant has 12 px.
https://www.figma.com/file/rWXEQiJcBXbkMY8V4NSM1d/ZenGRC-Component-Library?node-id=2629%3A3519

src/components/zen-avatar-details/zen-avatar-details.stories.mdx
```diff
@@ -20,9 +21,63 @@ export const StoryWithControls = args => {
...
+      <sb-zen-space vertical spacing="md" horizontal-align="center">
+        <sb-zen-avatar-details
+          id="avatar-basic-medium"
+          variant="basic"
+          style="border: 1px solid lightgray"
+        ></sb-zen-avatar-details>
+        <sb-zen-text size="sm">Variant 'Basic' medium</sb-zen-text>
+      </sb-zen-space>
+      <sb-zen-space vertical spacing="md" horizontal-align="center">
+        <sb-zen-avatar-details
```
chavlji, line 56: Change to `lg` size. Designs provide only `md` and `lg` size.

src/components/zen-avatar-details/zen-avatar-details.stories.mdx
```diff
@@ -20,9 +21,63 @@ export const StoryWithControls = args => {
...
+};
+
 # Avatar icon, username and email
 
-Avatar details component
+`Avatar details` component.
+
+<br />
+
+### 1. Avatar details variants
```
chavlji, line 74: NIT: Can be just simply "Variants"

chavlji (changes requested):
A few minor changes...

src/components/zen-avatar-details/zen-avatar-details.tsx
mihaelpalfi: Done

src/components/zen-avatar-details/zen-avatar-details.tsx
mihaelpalfi: Done

src/components/zen-avatar-details/zen-avatar-details.tsx
```diff
@@ -10,6 +10,12 @@ import { PaddingShorthand, Size } from '../helpers/types';
 export class ZenAvatarDetails {
   @Element() host: HTMLZenAvatarDetailsElement;
 
+  /** Different variants  */
+  @Prop() readonly variant: AvatarDetailVariant = 'detailed';
+
+  /** Sizes that apply only for variant basic  */
+  @Prop() readonly size: IconSizes = 'md';
```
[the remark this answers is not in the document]
mihaelpalfi: In my opinion would not go into direction what designers want regarding this case regarding naming and possibilities. Consumer should be able to chose from possible icon sizes.

src/components/zen-avatar-details/zen-avatar-details.stories.mdx
mihaelpalfi: Done

src/components/zen-avatar-details/zen-avatar-details.tsx
mihaelpalfi: This inconsistency is weird. Added a question for Shelley on wire frames if this is intentional.

src/components/zen-avatar-details/zen-avatar-details.tsx
mihaelpalfi: Good suggestion. I changed it. But not quite sure if i changed it in a correct way.

src/components/zen-avatar-details/zen-avatar-details.tsx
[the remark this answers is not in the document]
mihaelpalfi: Removed

src/components/zen-avatar-details/zen-avatar-details.tsx
mihaelpalfi: Shelley answered that it was a mistake. Its the same spacing for both. She updated the wireframes.

src/components/zen-avatar-details/zen-avatar-details.stories.mdx
mihaelpalfi: Removed size. Now we have three variants. Detailed, basic default and basic large.

src/components/zen-avatar-details/zen-avatar-details.tsx
```diff
@@ -33,8 +36,29 @@ export class ZenAvatarDetails {
   /** Skipped */
   @Prop() readonly paddingLeft: Spacing = null;
 
-  /** Spacing between icon and username  */
-  @Prop() readonly spacing: Size | None = 'md';
+  getPropValueByVariant(propertyName: string): string {
+    const propsByVariant = {
```
homer0, line 40: 1. This could be class property: since it doesn't change, we don't need to re define it on each call.
2. Instead of calling the function for each property, we could probably define a type for the dictionary:

```
type VariantSizes = {
  verticalAlignment: string;
  ...
}

...

getSizesByVariant(): VariantSizes { ... }
```

src/components/helpers/types.ts
```diff
@@ -27,6 +27,8 @@ export type Resize = 'none' | 'both' | 'horizontal' | 'vertical' | 'initial' | '
 
 export type PositionVariant = 'top-left' | 'top-right' | 'bottom-left' | 'bottom-right';
 
+export type AvatarDetailVariant = 'basic_default' | 'basic_large' | 'detailed';
```
homer0, line 30: Is there a reason we are using _snake_case_ for this instead of _dash-case_?

mihaelpalfi: Cant create object name with dash-case.

src/components/zen-avatar-details/zen-avatar-details.tsx
mihaelpalfi: Done

src/components/helpers/types.ts
homer0: But you can do it with quotes:

```js
'basic-default': { verticalAlignment: 'center', avatarIconSize: 'sm', userNameBold: false, textSize: 'md' }
```

I don't like it either, but I think it's a small price to pay for consistency :P.

src/components/zen-avatar-details/zen-avatar-details.tsx
```diff
@@ -49,22 +74,29 @@ export class ZenAvatarDetails {
           padding-right={this.paddingRight}
           padding-bottom={this.paddingBottom}
           padding-left={this.paddingLeft}
-          vertical-align="middle"
-          spacing={this.spacing}
+          spacing="md"
+          vertical-align={this.getPropsByVariant().verticalAlignment}
```
homer0, line 78: Why don't you call it only once and keep the reference?

```js
const sizes = this.getPropsByVariant();

...

<ZenSpace
  ...
  vertical-align={sizes.verticalAlignment}
>
```

homer0 (approved):
Tiny changes, but this is looking good 🤘!

src/components/zen-avatar-details/zen-avatar-details.tsx
mihaelpalfi: Done

src/components/helpers/types.ts
mihaelpalfi: Done.

src/components/helpers/types.ts
```diff
@@ -29,6 +29,8 @@ export type Resize = 'none' | 'both' | 'horizontal' | 'vertical' | 'initial' | '
 
 export type PositionVariant = 'top-left' | 'top-right' | 'bottom-left' | 'bottom-right';
 
+export type AvatarDetailVariant = 'basic-default' | 'basic-large' | 'detailed';
```
chavlji, line 32: nit: Please rename to `basic`, `basic-lg` and `detailed`

chavlji (changes requested):
Very clean PR! Great!

Just one minor thing to be more consistent...

src/components/helpers/types.ts
mihaelpalfi: Done.

chavlji (approved):

PR #266, opened 2021-03-25
QUICKFIX: fix lint:all and format:all scripts

homer0:
Follow-up for #263 - Both commands were calling Prettier :P

I fixed it so `lint:all` will call `eslint` and I changed both so they'll reuse the `lint` and `format` scripts `lint-staged` uses.


chavlji (approved):
Nice! Thanks!

PR #268, opened 2021-03-26
PLAT-1594: Add padding to rest of the components

chavlji:
[JIRA ticket](https://reciprocitylabs.atlassian.net/browse/PLAT-1594)


mihaelpalfi (changes requested):
Checked components. Looks ok. Only found that on skeleton padding is not applied and on avatar there is a missing  padding="none" prop for avatar details comp.

mihaelpalfi (approved):
Works. NIT : The separator line is not the same as in designs.
image

PR #275, opened 2021-03-31
PLAT-1548: Change properties to be reflective

mihaelpalfi:
[JIRA ticket](https://reciprocitylabs.atlassian.net/browse/PLAT-1548)

#### Description
As per JIRA ticket

#### ChangeLOG

#### Useful links


src/components/zen-table-header-cell/test/zen-table-header-cell.spec.tsx
```diff
@@ -16,7 +16,6 @@ describe('zen-table-header-cell', () => {
       components: [ZenTableHeaderCell],
       html: `<zen-table-header-cell sticky>Content</zen-table-header-cell>`,
     });
-
-    expect(page.root.getAttribute('class')).toEqual('sticky');
+    expect(page.root.hasAttribute('sticky')).toBeTruthy();
```
chavlji, line 19: Remove this test. If you set attr in the markup, it will have it...
You could add header-row test where you'd check if all children gets attr when you set header.sticky. (but it's a bit tricky. Not sure if this.host.children will work in jest since it does't support slot assignedNodes()... if it's too hard, don't bother)

src/components/zen-table-row/zen-table-row.tsx
```diff
@@ -164,15 +164,8 @@ export class ZenTableRow {
...
-    const hostClass = {
-      hidden: !this.visible,
-      selectable: this.selectable,
-      expandable: this.expandable,
-      selected: this.selected,
-      expanded: this.expanded,
-    };
     return (
-      <Host class={hostClass}>
+      <Host>
```
chavlji, line 168: Awesome simplification! 😎

chavlji (changes requested):
Awesome! One minor thing...

src/components/zen-table-header-cell/test/zen-table-header-cell.spec.tsx
mihaelpalfi: Removed. And the test for checking if child elements get the sticky class already exists in table header comonent.

chavlji (approved):

PR #279, opened 2021-03-31
QUICKFIX: Enable NPM && Fix Storybook deploy

homer0:
- Now that we know that the deploy works, I re-enabled NPM deploys.
- Instead of generating the Storybook on the `build` step, since it needs to be updated after `semantic-release`, I moved it to the `publish` step and made the `build` step run only Stencil.

> Yes, the commit to enable NPM should be a `chore:`, but I want to trigger a release, that's why I used `fix:`.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -91,11 +91,10 @@ jobs:
       - checkout
       - attach_workspace:
           at: .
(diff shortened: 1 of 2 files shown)


leferreyra (approved):
🚀

PR #280, opened 2021-03-31
PLAT-1611: inputs: add size prop

chavlji:
[JIRA ticket](https://reciprocitylabs.atlassian.net/browse/PLAT-1611)

I'm sorry that pr is quite big. I should've created a separate pr for each component, but... 🙄 

Note that the height of components changed a bit (for pixel or two), but imo it's not a breaking change.


src/components/zen-dropdown/test/zen-dropdown.spec.tsx
```diff
@@ -47,13 +50,15 @@ describe('zen-dropdown', () => {
     await page.waitForChanges();
     const focusin = new Event('focus', { bubbles: true, composed: true });
     dropdown.dispatchEvent(focusin);
-  });
+  };
 
   it('should open on focus', async () => {
+    await render();
```
homer0, line 56: Why not moving it to the `beforeEach`?

src/components/zen-dropdown/zen-dropdown.stories.mdx
```diff
@@ -159,16 +193,27 @@ A standard dropdown component
...
+
+<Canvas>
+  <Story name="StorySizes">{StorySizes.bind({})}</Story>
+</Canvas>
+
+### 2. Options with custom content
 
 - Each option has an icon and custom padding.
 - Because option is taller than usually, we have to specify dropdown's height manually, to match the option height.
+- Set `size` to `custom` in order. It will make dropdown's height equal to height selected option.
```
homer0, line 210: I think there's something missing here "...to `custom` in order...", in order to do what? :P

src/components/zen-dropdown/zen-dropdown.tsx
```diff
@@ -271,6 +275,8 @@ export class ZenDropdown {
         break;
     }
 
+    const iconSize = this.size === 'sm' ? 'sm' : 'md';
```
homer0, line 278: Ha! this is what I suggested on the previous comment xD.

src/components/zen-menu-item/zen-option.stories.mdx
```diff
@@ -50,6 +43,55 @@ export const Slots = args => {
...
+
+Option item to be used inside **zen-dropdown** component
+
+## Component previews
+
+<br />
+
+### 1. Sizes
+
+- Zen-option ships in different sizes
```
homer0, line 85: I've seen that you used this phrase in multiple places, and I think it doesn't fully _transmit_ what you are trying to say. What do you think about doing something like this:

```
This component supports multiple sizes through the use of the `size` attribute, but it is only considered when implemented with "default items" (simple text), and in that case, you need to make sure you set the same `size` on the `zen-dropdown`.
```

I know this can't be applied the same way on all the cases, but I just wanted to give you an example of how to phrase it like a message rather than a requirement list... and yes, I know that a requirement list for a library makes more sense, but it's a lot easier for the implementation when the connection between the "requirements" is explicit.

Anyways, let know what do you think :D

homer0 (commented):
This is looking great 🤘!

src/components/zen-date-picker/zen-date-picker.tsx
```diff
@@ -190,6 +194,10 @@ export class ZenDatePicker {
     }
   }
 
+  inputIconSize(): string {
+    return this.size === 'sm' ? 'sm' : 'md';
+  }
+
```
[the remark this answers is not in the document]
chavlji: Good points. Fixed.
Initially, I was thinking that testing will be easier with a separate method, but since it's not public it's not easier 😛

src/components/zen-dropdown/test/zen-dropdown.spec.tsx
chavlji: Because I send initial params as args to render function. I can't do that elegantly with beforeEach (not without global var hack).

mihaelpalfi (changes requested):
Looking fantastic. Tested on firefox and edge. Found only two issues : 
1. In Firefox story title overlaps opened dropdown.
image

2. When I set menu width to 50% the size of menu is not correct.
image

chavlji:
@mihaelpalfi Thanks. Fixed!

mihaelpalfi (approved):
LGTM

PR #288, opened 2021-04-01
PLAT-1696: Added focusInput method to zen-textarea

f-jauregui:
[JIRA ticket](https://reciprocitylabs.atlassian.net/browse/PLAT-1696)

#### Description
As per JIRA ticket

#### ChangeLOG

#### Useful links

### DIFF
--- a/src/components/zen-textarea/zen-textarea.tsx
+++ b/src/components/zen-textarea/zen-textarea.tsx
@@ -14,7 +14,7 @@ import { Resize } from '../helpers/types';
   shadow: true,
 })
 export class ZenTextarea {
-  input = null;
+  input: HTMLTextAreaElement | null = null;
 
   @Element() host: HTMLZenTextareaElement;
 
@@ -48,7 +48,9 @@ export class ZenTextarea {
   /** Focus input */
   @Method()
   async focusInput(): Promise<void> {
-    this.input.focus();
+    if (this.input) {
+      this.input.focus();
+    }
   }
 
   private onInput = (ev: Event) => {
--- a/src/components/zen-textarea/zen-textarea.tsx
+++ b/src/components/zen-textarea/zen-textarea.tsx
@@ -1,4 +1,4 @@
-import { Component, Host, h, Prop, Element } from '@stencil/core';
+import { Component, Host, h, Prop, Element, Method } from '@stencil/core';
 import { Resize } from '../helpers/types';
 
 /**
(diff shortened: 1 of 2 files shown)


src/components/zen-textarea/zen-textarea.tsx
```diff
@@ -14,6 +14,8 @@ import { Resize } from '../helpers/types';
   shadow: true,
 })
 export class ZenTextarea {
+  input = null;
```
homer0, line 17: Shouldn't this be `input: HTMLTextAreaElement | null = null`?

src/components/zen-textarea/zen-textarea.tsx
```diff
@@ -43,6 +45,12 @@ export class ZenTextarea {
   /** Prefilled text content */
   @Prop({ mutable: true }) text?: string | null = '';
 
+  /** Focus input */
+  @Method()
+  async focusInput(): Promise<void> {
+    this.input.focus();
```
homer0, line 51: And since `input` is nullable, you should always validate it before calling it:

```js
if (this.input) {
  this.input.focus();
}
```

src/components/zen-textarea/zen-textarea.tsx
f-jauregui: Yes, updated!

src/components/zen-textarea/zen-textarea.tsx
```diff
@@ -43,6 +45,14 @@ export class ZenTextarea {
   /** Prefilled text content */
   @Prop({ mutable: true }) text?: string | null = '';
 
+  /** Focus input */
+  @Method()
+  async focusInput(): Promise<void> {
```
mihaelpalfi, line 50: NIT: Do you think this method could be called only focus. Dont know what else should be focused in the textarea?

mihaelpalfi (approved):
Works. Just a thought. Should the focus event be propagated like we do it for change on ex. dropdown?

homer0 (approved):
🤘

src/components/zen-textarea/zen-textarea.tsx
f-jauregui: No, because the textarea already has a focus() method in its prototype.

homer0:
@mihaelpalfi To your questions, if you mean triggering a new event, I would validate that on a TD ticket, otherwise, the scope of the PR would change from "adding a missing method" to "fixing two components" (because if we are doing it here, we should do it for the input too)

PR #294, opened 2021-04-05
PLAT-1607: Remove size lg from avatar-icon

f-jauregui:
[JIRA ticket](https://reciprocitylabs.atlassian.net/browse/PLAT-1607)

#### Description
As per JIRA ticket

#### ChangeLOG

#### Useful links


mihaelpalfi (commented):
Good work. But i think you should not remove size lg from type IconSize. It is use also for the zen-icon component and from the wireframes the icons has 3 sizes.
image

Maybe it would be better to create a new type like AvatarIconSize...

f-jauregui:
> Good work. But i think you should not remove size lg from type IconSize. It is use also for the zen-icon component and from the wireframes the icons can have 3 sizes.
> image
> 
> Maybe it would be better to create a new type like AvatarIconSize...

Good catch, @mihaelpalfi. I'll correct that.

homer0 (approved):
🤘

Not a blocker, but I would recommend you to avoid plural names for union types; the type is one, the values are multiples.

mihaelpalfi (approved):
Good work. Just change the types to singular (AvatarIconSize and IconSize).

PR #297, opened 2021-04-06
PLAT-1627: Table component full span elements overlap issue

chavlji:
[JIRA ticket](https://reciprocitylabs.atlassian.net/browse/PLAT-1627)

#### Description
As per JIRA ticket

Edit: Unit tests are semi-fixed... `try catch` added so error is handled. Couldn't prevent it in the first place...


mihaelpalfi (approved):
Fantastic research regarding the grid and implementation Jan! This one is a true team effort :). Tested on firefox and ie and have a couple of things i noticed. All non blockers and can be done in separate prs : 
1. On the row story if i check is header and selected, it dosent select it.
2. Why is there an additional cell below on row story 3?
image
3. On sticky header story the header jumps a little when scrolling.
4. I know this is not part of this pr. But its just a small thing, in the row story please change <docs-table> to <sb-docs-table>

chavlji:
💪🏻 🚀 

1. Header can only be selected if all table rows are selected. And in that case we have no row in this table.
2. Fixed
3. That's Chrome's sticky bug. It was the same before. Nothing to do about it.
4. Fixed

homer0:
Ok, I tried debugging the issue with Jest and... I couldn't fix it 😞 , but I found some clues:

The way you are mocking `getDefaultSlotContent` doesn't work as you would expect: updating the reference of the variable doesn't update the return value of the mock, so the first time it runs, the mock always returns the empty array; that's why it throws the errors when you run more than one case: the errors only show up when there are actual cells to update the props.

The other thing I found is that by adding extra `await page.waitForChanges();`s I was able to reduce the amount of errors: since we have a lot of things going on when we change a prop, it seems like it's necessary to wait for the changes "before moving on". This example will explain it better:

```diff
 it('should set indeterminate state to all parent rows', async () => {
   const checkbox = secondDepthCell.shadowRoot.querySelector('.checkbox') as HTMLZenCheckboxElement;
   checkbox.checked = true;
+  await page.waitForChanges();
   const event = new Event('change', { bubbles: true, composed: true });
   checkbox.dispatchEvent(event);
   await page.waitForChanges();

   expect(parentRow.$indeterminate).toBeTruthy();
   expect(secondDepthParentRow.$indeterminate).toBeTruthy();
 });
```

I added it after the change of `checked` and after other props changes, and I got rid of the errors that show up before the cases list is printed on the terminal.

I would suggest we merge this PR like this, and we do a "test refactor" on a different one.

Amazing work 🤘

homer0 (approved):

chavlji:
@homer0 I'm merging it and I've created a new ticket to fix these test issues:
https://reciprocitylabs.atlassian.net/browse/PLAT-1794

PR #300, opened 2021-04-07
PLAT-1691: Updated zen-option mouseover hover style and remove keyboard focused style when mouseover

f-jauregui:
⚠️ Not sure I totally understood the issue in this ticket but let's see if this solve it.

[JIRA ticket](https://reciprocitylabs.atlassian.net/browse/PLAT-1691)

#### Description
As per JIRA ticket

#### ChangeLOG

#### Useful links

### DIFF
--- a/src/components/zen-menu-item/zen-option.scss
+++ b/src/components/zen-menu-item/zen-option.scss
@@ -44,18 +44,21 @@ $item-line-height: 1.2rem;
   line-height: 2.5rem;
 }
 
-:host(:not([no-hover]):hover),
-:host([selected]),
-:host([focused]) {
+:host([selected]) {
+  .background {
+    font-weight: bold;
+  }
+}
+
+:host(:not([no-hover])[focused]) {
   .background {
     background-color: $color-gray-100;
   }
 }
 
-:host([selected][focused]),
-:host(:not([no-hover])[selected]:hover),
-:host(:not([no-hover])[focused]:hover) {
+:host(:not([no-hover])[selected][focused]) {
   .background {
-    background-color: $color-gray-200;
+    background-color: $color-gray-100;
+    font-weight: bold;
   }
 }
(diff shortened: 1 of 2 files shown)


homer0:
I don't think this is working properly:

image

I believe that the scroll that happens when the focus changes, it's re-triggering the mouse event after you "unselected" all siblings.

I think the solution should be something like: "when you use the arrows, a flag that allows the style to change on mouse over is set to `false` and cannot be changed for ~1sec, and when there's a mouse move, the flag changes to `true`". Not sure if the the "waiting" is necessary, but if the scroll re triggers the mouse move, it would be the same situation.

chavlji:
Yeah. We should have private vars `keyboardFocusedItem` and `mouseFocusedItem`. They should mutually exclude:
- If user mouseover item, `keyboardFocusItem` is set to `null`.
- If `mouseFocusedItem` is set and user presses `down` key, `keyboardFocusedItem` becomes next item from `mouseFocusedItem` and `mouseFocusedItem` is set to null.
- If `mouseMove` or `mouseOver` occurs on item while `keyboardFocusedItem` is set, `keyboardFocusedItem` should be set to `null` and `mouseFocusedItem` is set to mouse hovering item.

homer0:
@f-jauregui I added the `WIP!` label, as the solution would be different as the one in the PR.

chavlji:
Nice! It works as expected!

One NIT: I get this linter complaints:
Screenshot 2021-04-12 at 16 09 57

chavlji (approved):
Nice! It works as expected!

One NIT: I get this linter complaints:
Screenshot 2021-04-12 at 16 09 57

homer0:
I hate to be that guy, but...there's still a bug :P:

https://user-images.githubusercontent.com/6949310/114429064-54cab280-9b93-11eb-9669-a00696dc5e2b.mov

If, for some reason, the mouse is over an item when you move with the keyboard, is that items enters the viewport again, it steals the focus from the keyboard. In the case of the video, I couldn't choose the first option with the keyboard.

chavlji:
Is this fixed? It looks to me that it's working as expected... what baffles me is that there's only 1 commit. Please add fixes after reviews as a separate commit so it's easier to review changes.

mihaelpalfi (approved):
Works as expected!

f-jauregui:
> @f-jauregui Is this fixed? It looks to me that it's working as expected... what baffles me is that there's only 1 commit. Please add fixes after reviews as a separate commit so it's easier to review changes.

Yes. Sorry about that, was mainly to keep the commit clean. But just changing the mouseover to mousemove did the trick.

PR #301, opened 2021-04-08
PLAT-1755: Add tabs component

mihaelpalfi:
[JIRA ticket](https://reciprocitylabs.atlassian.net/browse/PLAT-1755)

#### Description
As per JIRA ticket

#### ChangeLOG

#### Useful links


src/components/zen-tab/zen-tab.tsx
```diff
@@ -0,0 +1,41 @@
...
+  }
+
+  onClick(): void {
+    this.selected = !this.selected;
+    this.tabSelect.emit();
+  }
+
+  render(): HTMLZenTabElement {
+    return (
+      <Host class={{ selected: this.selected }} onClick={() => this.onClick()}>
```
chavlji, line 36: When you promote `selected` to prop, you should just reflect it instead of converting to class.

src/components/zen-tab/test/zen-tab.spec.tsx
```diff
@@ -0,0 +1,24 @@
...
+  it('should render', async () => {
+    const page = await newSpecPage({
+      components: [ZenTab],
+      html: `<zen-tab></zen-tab>`,
+    });
+
+    expect(page.root.shadowRoot).toBeTruthy();
+  });
+
+  it('should select', async () => {
```
chavlji, line 14: If you reflect `selected` you don't need to check this.

src/components/zen-tab/zen-tab.tsx
```diff
@@ -0,0 +1,41 @@
...
+  @Element() host: HTMLZenTabElement;
+
+  /** Is tab selected */
+  @State() selected = false;
+
+  /** Tab selected event */
+  @Event() tabSelect: EventEmitter<void>;
+
+  /** Deselect tab */
+  @Method()
```
chavlji, line 18: Promote `selected` to prop and set watcher for it. Then you can remove methods `select` and `deselect`. Components should "follow props down, events up" if possible.

chavlji (changes requested):
Looks nice and works as expected. One minor improvement though...

src/components/zen-tab/zen-tab.tsx
mihaelpalfi: done

src/components/zen-tab/test/zen-tab.spec.tsx
mihaelpalfi: done

src/components/zen-tab/zen-tab.tsx
mihaelpalfi: removed

chavlji (approved):
Looks great! 😎

homer0:
@mihaelpalfi Looks great... but :P:

1. I don't see any commit with the `BREAKING CHANGE:` comment, and the `zen-tabs` implementation is completely different now.
2. Can you change the PR target to the `breaking` branch? so we can release it today with the rest of the breaking changes.

mihaelpalfi:
> @mihaelpalfi Looks great... but :P:
> 
> 1. I don't see any commit with the `BREAKING CHANGE:` comment, and the `zen-tabs` implementation is completely different now.
> 2. Can you change the PR target to the `breaking` branch? so we can release it today with the rest of the breaking changes.

We can. But my opinion is that this tabs were in the "in progress" in the story book before and is not used nowhere right now. I have only one thing to push when its reviewed. Move it to the navigation part of the story book.

homer0 (approved):

homer0:
@mihaelpalfi Ok, that's fine for this case, but we shouldn't publish stuff in which we are not going to respect SemVer; the version/changelog is what should tell implementation whether or not something changed, not a section of Storybook.

Also, not a blocker (I already sent the approval), but shouldn't the tabs component emit a change event? Otherwise, the implementation would have to add `onClick` on each tab.

chavlji:
@mihaelpalfi can we merge this?

PR #302, opened 2021-04-08
PLAT-1772: Fix for initials not displayed correctly

mihaelpalfi:
[JIRA ticket](https://reciprocitylabs.atlassian.net/browse/PLAT-1772)

- Tried all the different regex expressions (also suggested ones) and could not make it work. :(
- Added override option for initials to all needed components.

#### Description
As per JIRA ticket

#### ChangeLOG

#### Useful links

### DIFF
--- a/src/components/zen-avatar-details/zen-avatar-details.tsx
+++ b/src/components/zen-avatar-details/zen-avatar-details.tsx
@@ -31,7 +31,7 @@ export class ZenAvatarDetails {
 
   @Element() host: HTMLZenAvatarDetailsElement;
 
-  /** Set fifferent variants  */
+  /** Set the style variant  */
   @Prop() readonly variant: AvatarDetailVariant = 'detailed';
 
   /** Set user name  */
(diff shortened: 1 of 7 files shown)


src/components/zen-avatar-details/zen-avatar-details.tsx
```diff
@@ -31,19 +31,22 @@ export class ZenAvatarDetails {
 
   @Element() host: HTMLZenAvatarDetailsElement;
 
-  /** Different variants  */
+  /** Set fifferent variants  */
```
homer0, line 34:
```suggestion
  /** Set the style variant  */
```

First I wanted to change the `fifferent` :P, but then I realized that the implementation doesn't set "variants", it chooses a single variant from a list.

homer0 (approved):
image

🤘

src/components/zen-avatar-details/zen-avatar-details.tsx
mihaelpalfi: Good catch. And i like this fix suggestions!

PR #304, opened 2021-04-09
PLAT-1796: Table: Make sure table structure and state is always correct

chavlji:
[JIRA ticket](https://reciprocitylabs.atlassian.net/browse/PLAT-1796)

1st part of ticket [PLAT-1667: Table: indeterminate state inconsistent](https://reciprocitylabs.atlassian.net/browse/PLAT-1667)

#### Description
As per JIRA ticket

#### Solution:

The idea is, to have one **central cleanup function** `cleanupTableStructure` that will check the table each time something is changed and it will maintain the correct table structure. This approach seems the most straightforward (therefore the easiest to maintain) and it should cover all edge-cases that would otherwise need to be handled manually per case.

This function should:

keep selected/indeterminate state of parent rows

clean up the nested structure (based on row.depth) and make sure there are no inaccessible rows

update item’s visibility (based on parent.expanded)

**Central cleanup function** `cleanupTableStructure` should be called by table component (since all other components here are table’s children). Table should use mutation observer to monitor all it’s children and run the function on any child change. Mutation observer triggers callback in bulks, so there should be no performance issues.

I’ve added a simple table editor to insert/remove rows. Play with it and make sure table state is as expected in any case:

https://user-images.githubusercontent.com/5729421/114175175-fce53f00-9939-11eb-80b3-a16d3559afe5.mov


f-jauregui (approved):
What an awesome job, Jan! 🥇 
Love that observer and that updates everything in bulk.

mihaelpalfi (changes requested):
Great and fast work Jan. And fantastic idea with table editor! Focused more on testing the functionality.  Found some issues
1.  With the header row transition from indeterminate state to checked (chrome).

https://user-images.githubusercontent.com/59830865/114190113-396d6680-994b-11eb-8eeb-cf586eeff19f.mov

2. Probably the same issue when deselecting the second level child the state of main parent checkbox dosent change. 

https://user-images.githubusercontent.com/59830865/114189981-180c7a80-994b-11eb-86fd-07465d373da9.mov

src/components/zen-table/test/zen-table.spec.tsx
```diff
@@ -1,7 +1,16 @@
+import { mutationObserverMock } from '../../helpers/jest';
+global.MutationObserver = mutationObserverMock();
+
+global.console = { error: jest.fn() };
```
homer0, line 4: You should use a spy instead of overwriting the global console, otherwise you would kill the functionality even for the runner:

```js
const spy = jest.spyOn(console, 'error');
```

And since you are only validating it on a single test, I would suggest that you create the spy only in that test.

src/components/helpers/helpers.ts
```diff
@@ -128,3 +128,9 @@ export function scrollIntoView(element: HTMLElement, scrollParent: HTMLElement):
     scrollParent.scrollBy(0, bottomDistance);
   }
 }
+
+export function htmlToElement(html: string): Element[] {
```
homer0, line 132: Why are we shipping stuff that's only meant for the tests and Storybook? My concern is that, if we ship it, we have to maintain it, meaning that changes on this function will affect the versioning of the package, when we don't actually expect for implementations to use it.

src/components/zen-table/test/zen-table.spec.tsx
```diff
@@ -1,7 +1,16 @@
+import { mutationObserverMock } from '../../helpers/jest';
+global.MutationObserver = mutationObserverMock();
```
homer0, line 2: If there's no other way and you have to overwrite a global object, what you normally do is: you keep a reference of the original, replace it on `beforeEach` and restore it on `afterEach`, that way, the tests don't affect any other tests/functionality that may use/check the property:

```js
import { mutationObserverMock } from '../../helpers/jest';
const originalMutationObserver = global.MutationObserver;

describe('zen-table', () => {
  beforeEach(() => {
    mutationObserverMock.mockClear();
    global.MutationObserver = mutationObserverMock();
  });
  
  afterEach(() => {
    global.MutationObserver = originalMutationObserver;
  });
});
```

The idea is that tests cases should, when possible, avoid at all costs to modify the environment outside their execution.

src/components/zen-table/zen-table-helpers.tsx
```diff
@@ -0,0 +1,82 @@
+export const cleanupTableStructure = async function (table: HTMLZenTableElement): Promise<void> {
+  function parentRows(rows, index: number): HTMLZenTableRowElement[] {
```
homer0, line 2: - Why did you define other functions inside `cleanupTableStructure`? None of them use the `table` parameter. They could be non-exported functions.
- I would recommend you that, if you don't need to use `this`, you should use an arrow function. It's just a little, but arrow functions can generate a smaller "footprint" on the generated bundle, as they can be optimized in ways regular functions can't (like removing parentheses, brackets and/or the `functions and `return` keywords. The only difference is that arrow functions don't have context

src/components/zen-table/zen-table-helpers.tsx
```diff
@@ -0,0 +1,82 @@
+export const cleanupTableStructure = async function (table: HTMLZenTableElement): Promise<void> {
```
homer0, line 1: Is there a need for this to be `async`? I don't see any use of promises/`await`.

src/components/zen-table/zen-table-helpers.tsx
```diff
@@ -0,0 +1,82 @@
...
+      if (rows[i].depth <= myDepth) break;
+      children.push(rows[i]);
+    }
+    return children;
+  }
+
+  function removeOrphans(rows) {
+    // orphans: rows whose depth is more than 1 greater from prevSibling.depth
+    // rows like this could never become visible.
+    // Apprently their parent got deleted so we'd expect they'd get too.
```
homer0, line 26: This is very informative, so maybe you should move it to the JSDoc block for the function, that way IDEs can pick it.

src/components/zen-table/zen-table-helpers.tsx
```diff
@@ -0,0 +1,82 @@
...
+  // NOTE: Order of below function calls is important!!!
+  removeOrphans(rows);
+  updateExapndableProps(rows);
+  updateVisibleProps(rows);
+  updateParentCheckboxes(rows);
+
+  table.$updating = false;
+};
+
+export const allRows = function (table: HTMLZenTableElement): HTMLZenTableRowElement[] {
```
homer0, line 77:
```suggestion
export const getAllRows = function (table: HTMLZenTableElement): HTMLZenTableRowElement[] {
```

src/components/zen-table/zen-table-helpers.tsx
```diff
@@ -0,0 +1,82 @@
...
+  function isNormalRow(element: HTMLZenTableRowElement) {
```
homer0, line 78: Like above, no need to re define the function on each call if there's no use of the context:

```js
const isNormalRow = (element: HTMLZenTableRowElement): boolean => element.tagName.endsWith('-ROW') && !element.header;
export const getAllRows = (table: HTMLZenTable): HTMLZenTableRowElement[] => (Array.from(table.children) as HTMLZenTableRowElement[]).filter(isNormalRow);
```

src/components/zen-table/zen-table.tsx
```diff
@@ -1,20 +1,42 @@
 import { h, Component, Host, Element, Prop } from '@stencil/core';
+import { cleanupTableStructure } from './zen-table-helpers';
 
 @Component({
   tag: 'zen-table',
   styleUrl: 'zen-table.scss',
   shadow: true,
 })
 export class ZenTable {
+  childObserver: MutationObserver = null;
```
homer0, line 10:
```suggestion
  private childObserver: MutationObserver = null;
```

homer0 (commented):
image

I just left a couple of syntax-based comments :P.

Also, I believe the issues @mihaelpalfi reported are fixed on #305 , as I just tested it and I couldn't reproduce the erros from the videos.

src/components/zen-table/test/zen-table.spec.tsx
chavlji: If I don't overwrite it, we get console.error in the output which I think will crash tests on the ci:
Screenshot 2021-04-12 at 09 18 11
I've ended up mocking an unlocking `console.error` on each test.

src/components/zen-table/zen-table-helpers.tsx
chavlji: - I always do that. I think it's more clear this way, these functions are (for now) only used inside this one function.
- Until now I always did right the opposite. I only used arrow functions if I needed to persist the scope of the "parent" function. It's also easier to extract function this way if we also need it somewhere else later.
  - I've changed it now and will try to follow your pattern.

src/components/helpers/helpers.ts
chavlji: I'm opening a new ticket for that. How do we exclude it from bundle?
https://reciprocitylabs.atlassian.net/browse/PLAT-1811

chavlji:
@mihaelpalfi First issue should be gone in https://github.com/reciprocity/zen-ui/pull/305
The second issue isn't an issue. In this case, only the direct child has all children selected, so only this one will get checked checkbox. Other parents should have an indeterminate checkbox as they do.

mihaelpalfi (approved):

src/components/zen-table/zen-table-helpers.tsx
homer0: Not a big deal, but functions shouldn't be used for scoping, as you would end up re defining them each time it gets called, and in this case, the function is called pretty often.

If non-exported functions are not your thing, you could use any of these:

- Modules: Yes, you still have non-exported functions, but with a single context; something like `zen-table/helpers/cleanupTableStructure` and only export the main function.
- Static class: you can use methods (even private methods) for the utility functions. The only "downside" is that if you want to export only one method, you have to bind it, for example: `export const cleanTableStructure = Cleaner.clean.bind(Cleaner)` (where `Cleaner` is the name of the class).

PR #306, opened 2021-04-11
PLAT-1773: When clear button is clicked on zen-input a change event is triggered

f-jauregui:
[JIRA ticket](https://reciprocitylabs.atlassian.net/browse/)

⚠️ **NOTE: Nor onClick and onMouseup worked on this component, not sure if it's a Stencil bug.** ⚠️ 

#### Description
As per JIRA ticket

#### ChangeLOG

#### Useful links

### DIFF
--- a/src/components/zen-date-picker/tests/zen-date-picker.spec.tsx
+++ b/src/components/zen-date-picker/tests/zen-date-picker.spec.tsx
@@ -23,7 +23,6 @@ describe('zen-date-picker', () => {
   let datepicker: HTMLZenDatePickerElement;
   let input: HTMLZenInputElement;
   let calendar: HTMLZenPopoverElement;
-  let xButton: HTMLZenPopoverElement;
 
   const render = async (attributes: string) => {
     jest.clearAllTimers();
@@ -38,7 +37,6 @@ describe('zen-date-picker', () => {
     datepicker = page.root as HTMLZenDatePickerElement;
     input = datepicker.shadowRoot.querySelector('#date-input');
     calendar = datepicker.shadowRoot.querySelector('.calendar');
-    xButton = datepicker.shadowRoot.querySelector('.icon.clear');
   };
 
   async function focusInput() {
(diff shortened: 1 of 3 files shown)


chavlji (changes requested):
Nice. Two comments:
1. `input` event should be fired as well. As by definition
2. _"The input event fires when the value of an <input>, <select>, or <textarea> element has been changed."_ - from [mdn](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/input_event))
2. On date picker we get 2 change events on clear click. Could you please investigate why?

mihaelpalfi (approved):
Looks good. Change event triggered. Can confirm two events fired in date picker. I assume the issue is that it is logged for the input component and the date picker component. So story book.

f-jauregui:
@chavlji I asked for your review again because as @mihaelpalfi said, I think the double change event is caused by Storybook.

chavlji (approved):
Ok. If you have some time, please investigate why it's called twice in the storybook.

homer0:
It shouldn't have an `input` event; `input` is for user interaction, changing the value programmatically doesn't trigger it. This button is similar to `<input type="reset" />`, and if you listen for `input` and `change` events on an input that can be reset, the only triggered event is `change`.

Regarding the double event, is it possible that they are the `change` from the input and the `change` triggered from picker itself?

f-jauregui:
> It shouldn't have an `input` event; `input` is for user interaction, changing the value programmatically doesn't trigger it. This button is similar to `<input type="reset" />`, and if you listen for `input` and `change` events on an input that can be reset, the only triggered event is `change`.
> 
> Regarding the double event, is it possible that they are the `change` from the input and the `change` triggered from picker itself?

@chavlji @homer0 Should I roll back the last commit?

src/components/zen-date-picker/tests/zen-date-picker.spec.tsx
```diff
@@ -192,11 +181,6 @@ describe('zen-date-picker', () => {
     expect(JSON.stringify(datepicker.value)).toEqual('null');
   });
 
-  it('should not render x button if allowEmpty is false', async () => {
```
chavlji, line 195: I'd still test this? Test just if zen-input has attr `clear-button` set. Jic someone messes up your attr binding...

chavlji (approved):
Nice! Works as expected.
I'd just put back this test and then you can merge it! 🚢

src/components/zen-date-picker/tests/zen-date-picker.spec.tsx
homer0: Agree, even if the button is on another component, we should test that the attr set on the picker reflects correctly on the input attrs.

homer0 (approved):

PR #308, opened 2021-04-12
QUICK-FIX: item of vertical zen-space should not have right margins!

chavlji:
Example of a bug:
If zen-space is declared like this:
Screenshot 2021-04-12 at 08 45 45
We get the following result:
Screenshot 2021-04-12 at 08 45 23

Shouldn't have this spacing on the right!


mihaelpalfi (approved):
Works. Thank you for quick fix Jan!

PR #309, opened 2021-04-12
PLAT-1735: Storybook: Mark $props as readonly in props table

chavlji:
[JIRA ticket](https://reciprocitylabs.atlassian.net/browse/PLAT-1735)

**!WARNING!**
Based on https://github.com/reciprocity/zen-ui/pull/308! It should be merged first!

#### Description
As per JIRA ticket


.storybook/preview.scss
```diff
@@ -100,12 +100,36 @@ h1.sbdocs-h1 {
...
+    font-style: italic;
   }
 
-  tr:nth-child(2n - 1) {
-    background: rgba(0, 0, 0, 0.02) !important;
+  .readonly-prop::after {
+    content: 'READ ONLY';
+    font-style: normal;
+    position: absolute;
+    left: 2rem;
```
homer0, line 123: Suggestion: `1.25rem` - It aligns with the prop name :P. I don't know if you were going for the indentation, that's why it's a suggestion xD.

.storybook/helpers/argTypes.js
```diff
@@ -112,8 +121,18 @@ export function spreadArgs(args, argTypes) {
...
   }
 
+  const propsToAttributes = (args, argTypes) => {
+    const attrs = {};
+    for (const [prop, value] of Object.entries(args)) {
+      const attr = argTypes[prop] ? argTypes[prop].attr : prop;
+      attrs[attr] = value;
+    }
+    return attrs;
+  }
```
homer0, line 131: Not a blocker, but you could do it smaller and a single liner with a reducer:

```js
const propsToAttributes = Object.entries(args).reduce(
  (acc, [prop, value]) => ({
    ...acc,
    [argTypes[prop] ? argTypes[prop].attr : prop]: value,
  }),
  {}
);
```

homer0 (approved):
Looks great 🤘

.storybook/helpers/argTypes.js
chavlji: Changing to this one, but are you sure this is more readable? 😊 

Btw (this doesn't apply here, but I'm just curious...)  in terms of optimization, for loop would be faster right? Since reducer involves a lot of object construction and destruction?

homer0: - With something that small, maybe the only thing I would do to make it more readable would be defining the `attr` var, like you were doing.
- You would be surprised, there's a lot of discussion around this kind of things that started when immutability became a good practice in JS; in terms of "speed", `for of` is faster, in terms of "memory management", I'm not entirely sure... the `for of` should be the winner, but there are some arguments about retaining references and garbage collection that would make you cry :P. My rule of thumb for micro optimization on JS loops is "if there are less than 1000 items, I won't even think about it" xD.

chavlji: Nice! Thanks for this! Yeah, It's completely worthless of thinking about optimization here since it happens once per page load! Even 1000 items seem like a low number I guess (except if it's in some raf function) :D

PR #312, opened 2021-04-12
PLAT-1270: Generate && Publish React bindings

homer0:
[JIRA ticket](https://reciprocitylabs.atlassian.net/browse/1270)

#### Description

There's a lot of code and a lot of commits, so let me tell you how it went:

Stencil can generate React bindings when building, but there are two caveats:

1. The generated code is TypeScript, so you need to "build the code you generated when you built Stencil". The way the tutorial goes, you have to copy the generated code into another repository (as @f-jauregui mentioned on the PR Matics did)... that looked unnecessary.
2. The generated code doesn't support tag names transformations... so the thing with the prefixes was an issue.

After building it manually for testing, I found out that some types were missing on the generated output, which meant that I couldn't run "the second build", that's why there's a `fix` commit (the one that will trigger the release).

Next step, I created a "base subproject" inside the `zen-ui` repository, with the necessary `package.json` and empty folders for placing the output, rebuilding it and publish, and that's the new `react` directory.

With the "subproject", I partially removed the necessity of other repository, but I still had the issue with the prefixes, and that's why I created the `transform.ts` script. It modifies the generated code in order to add some extra functionality, so we can send the same function we send to the `defineCustomElements` function.

> I was going to do AST parsing && update, but I've never done it with TypeScript code and it looked a little bit more complicated, so I used plain regexs for now.

On the file that defines generic types for the output, it adds these types:

```ts
export type TransformTagFn = (tagName:string) => string;
export interface LibConfig { transformTagName: TransformTagFn; };
```

On the entry file for the package, it adds this:

```ts
import type { TransformTagFn, LibConfig } from './react-component-lib/interfaces';

const CONFIG: LibConfig = {
  transformTagName: (tagName) => tagName,
};

export const setTagTransform = (fn: TransformTagFn) => {
  CONFIG.transformTagName = fn;
};
```

In the same file, we have the definition of the wrappers, which I changed to include `CONFIG`:

```diff
- export const ColorSwatch = /*@__PURE__*/createReactComponent<JSX.ColorSwatch, HTMLColorSwatchElement>( 'color-swatch');
+ export const ColorSwatch = /*@__PURE__*/createReactComponent<JSX.ColorSwatch, HTMLColorSwatchElement>(CONFIG, 'color-swatch');
```

Finally, on the file with the function that creates the wrapper:

```ts
import type { LibConfig } from './interfaces';
```

&&

```diff
export const createReactComponent = <
  PropType,
  ElementType extends HTMLStencilElement,
  ContextStateType = {},
  ExpandedPropsTypes = {}
>(
+ config: LibConfig,
  tagName: string,
  ReactComponentContext?: React.Context<ContextStateType>,
  manipulatePropsFunction?: (
    originalProps: StencilReactInternalProps<ElementType>,
    propsToPass: any,
  ) => ExpandedPropsTypes,
) => {
```

&&

```diff
- return React.createElement(tagName, newProps, children);
+ return React.createElement(config.transformTagName(tagName), newProps, children);
```

And with that, we now can import `setTagTransform` from the package and send the same function as on `DefineCustomElements` and it will work :D.

> Next revision will include a single function to do both things :D - I didn't want to "modify" too much code now.

Done with the transformation! Moving to the publishing part...

Since the main package is already wired with semantic-release, I couldn't update the script to publish multiple things on different processes... so I created a second script: `publish.ts`.

The script will read the name of the registries from the `.releaserc` we use for the main package, and get the env vars the plugin we use needs, but the publishing part is done manually:

1. It first validates that the last commits was a `chore(release)` as it should only publish the subproject when the main project was published.
2. Then, it modifies the subproject `package.json` in order to install `zen-ui` from the path. Since the publishing for `zen-ui` and the subproject will have almost at the same time, I wanted to avoid issues with the registry (as it sometimes takes a few seconds to show up new versions).
2. After installing the deps, it modifies the `package.json` again to point to the new version.
3. Builds the subproject using `tsc`.
4. Finally, for each of the registries, it modifies the `.npmrc` of the subproject with the info of the env vars, and publishes.

A few extra notes:

1. I added a couple of `react:` scripts on the main `package.json` that are either hooked to the build process or the deployment.
2. I updated the documentation for the getting started.

image


stencil.config.ts
```diff
@@ -8,7 +8,7 @@ export const config: Config = {
   outputTargets: [
     reactOutputTarget({
       componentCorePackage: packageName,
-      proxiesFile: './react/components.ts',
```
chavlji: I don't like putting auto-generated files into src... Is the purpose of that this folder is published and exposed in the final package's bundle?

src/stories/getting_started.stories.mdx
```diff
@@ -54,3 +54,42 @@ Read this before consuming these components.
...
+```
+
+1. Since using a prefix is required, you need to set it both for the library and the bindings.
+2. The reason you need to do it "twice" it's because `defineCustomElements` supports extra options that the bindings don't.
+
+Once the library and bindings are configured, you can just create a component and import the components:
+
+```jsx
+import React from 'react';
+import { ZenDropdown, ZenOption } from '@reciprocity/zen-ui-react';
```
chavlji, line 86: Awesome!!! 🚀

react/publish.ts
```diff
@@ -191,7 +224,8 @@ const prepareConfig = (config: Config): Config => ({
  */
 (async () => {
   const beforeTime = new Date().getTime();
-  const config = prepareConfig(CONFIG);
+  const config = await prepareConfig(CONFIG);
+  console.log({ config });
```
chavlji, line 228: NIT: Is this on purpose? Do we want to print the config?

chavlji (approved):
This looks really awesome! Great gob!!!🥇 🚀 

(I just went through commits - which are very nicely structured - and I haven't tried it on the React app)

react/publish.ts
homer0: This is called "it was too late, my wife was calling me, and I forgot to remove the debug line" :P

stencil.config.ts
homer0: Beware, you created the comment from an old commit, this path is now `./react/src/output/components.ts`, and everything inside `output` is not tracked.

The reason this goes inside `react/src` is that it's used to build the react bundle, but don't worry, this is not published with any of the packages. Both the main package and the react package use `files` in their `package.json` to specify which are the things that get published.

f-jauregui (approved):
```
THIS
IS
AWESOME
```

Nailed it the config, went through all your explanation step by step. It's a very hard task considering release timing and ordering above all. This will save us A LOT of time in the future. Fantastic work, man! 🎩 

Tested in a react app and components work perfectly with typings and all 😍

f-jauregui:
asd

PR #316, opened 2021-04-13
PLAT-1670: Remove unused helper classes

chavlji:
[JIRA ticket](https://reciprocitylabs.atlassian.net/browse/PLAT-1670)

#### Description
As per JIRA ticket


mihaelpalfi (approved):
Nice cleanup! And everything still works. From the failing snapshot test i think the issue is with the changes i made and changed the icon to an image. But i updated the snapshots and it should not fail for you. I can take a more deep look if you have problems resolving them.

chavlji:
@mihaelpalfi I've fixed it. Moved all storybook assets to /.storybook so they are no longer present in the bundle.

homer0 (approved):

PR #318, opened 2021-04-14
PLAT-1794: Test: Fix zen-table tests errors

chavlji:
[JIRA ticket](https://reciprocitylabs.atlassian.net/browse/PLAT-1794)

#### Description
As per JIRA ticket

### DIFF
--- a/src/components/zen-table-row/test/zen-table-row.spec.tsx
+++ b/src/components/zen-table-row/test/zen-table-row.spec.tsx
@@ -31,47 +31,9 @@ describe('zen-table-row', () => {
 
     expect(page.root.shadowRoot.querySelector('zen-checkbox')).toBeTruthy();
   });
-
-  it.skip('should render expendable icon and on click expand full span row', async () => {
-    const page = await newSpecPage({
-      components: [ZenTable, ZenTableRow, ZenTableCell],
-      html: /*html*/ `
-        <zen-table columns="1fr 1fr">
-          <zen-table-row depth="0">
-            <zen-table-cell class="first-cell">Row 1, Cell 1</zen-table-cell>
-            <zen-table-cell>Row 1, Cell 2</zen-table-cell>
-          </zen-table-row>
-          <zen-table-row full-span depth="1">
-            <zen-table-cell>Test</zen-table-cell>
-            <zen-table-cell></zen-table-cell>
-          </zen-table-row>
-        </zen-table>`,
-    });
-    const table = page.root as HTMLZenTableElement;
-    const mainRow = page.root.querySelector('zen-table-row') as HTMLZenTableRowElement;
-    helpers.getDefaultSlotContent = jest.fn(() => parentRow.querySelectorAll('zen-table-cell'));
-    mainRow.selectable = true;
-    await page.waitForChanges();
-    await cleanupTableStructure(table);
-
-    expect(mainRow.$visible).toBe(true);
-
-    const firstCell = page.root.querySelector('.first-cell') as HTMLZenTableRowElement;
-
-    const expandRow = page.root.querySelector('[full-span]') as HTMLZenTableRowElement;
-    expect(expandRow.$visible).toBeFalsy();
-
-    const expandableIcon = firstCell.shadowRoot.querySelector('.expand-icon');
-    expect(expandableIcon).toBeTruthy();
-
-    simulateMouse('click', expandableIcon);
-    cleanupTableStructure(table);
-    expect(mainRow.expanded).toBe(true);
-    expect(expandRow.$visible).toBe(true);
-  });
 });
 
-describe('zen-table-row tree functionality', () => {
+describe('zen-table-row inside tree structure', () => {
   let page: SpecPage;
   let table: HTMLZenTableElement;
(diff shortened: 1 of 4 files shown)


src/components/zen-table-row/test/zen-table-row.spec.tsx
```diff
@@ -108,13 +67,12 @@ describe('zen-table-row tree functionality', () => {
     });
     table = page.root;
     parentRow = table.querySelector('zen-table-row') as HTMLZenTableRowElement;
-    secondDepthParentRow = table.querySelector('[second-level-parent]') as HTMLZenTableRowElement;
+    secondDepthRow = table.querySelector('[second-level-parent]') as HTMLZenTableRowElement;
 
     // mock slots:
-    cells = parentRow.querySelectorAll('zen-table-cell');
+    helpers.getDefaultSlotContent = jest.fn(() => parentRow.querySelectorAll('zen-table-cell'));
```
mihaelpalfi, line 73: Code lint shows me this error TS2540: Cannot assign to 'getDefaultSlotContent' because it is a read-only property. I would not mention it, but i get this errors also when switching branches. And I have to restart the app so they go away.

mihaelpalfi (approved):
Nice looking test i must say. NIT: Looked at the coveralls and if you are up to it maybe a test for the row as header could be added? It would increase also the zen-table-helper coverage.

src/components/zen-table-row/test/zen-table-row.spec.tsx
```diff
@@ -1,18 +1,15 @@
...
-
+import * as helpers from '../../helpers/helpers';
+import { cleanupTableStructure } from '../../zen-table/zen-table-helpers';
 import { newSpecPage } from '@stencil/core/testing';
 import { ZenTableRow } from '../zen-table-row';
 import { ZenTable } from '../../zen-table/zen-table';
 import { ZenTableCell } from '../../zen-table-cell/zen-table-cell';
-import { cleanupTableStructure } from '../../zen-table/zen-table-helpers';
+
+global.MutationObserver = mutationObserverMock();
```
homer0, line 10: You should do here what you did for the `getDefaultSlotContent`: store the original and restore it after the tests; the reason is that `MutationObserver` is an object of the global context.

Now, TypeScript is going crazy with this :P - `mutationObserverMock` (I would add a `create` or `get` as a prefix) returns `unknown`, but you are using it as a mock; check [this](https://gist.github.com/homer0/916c92b3b603cbcbceda3e28d8854395), I added an interface and a couple of types so TS won't complain.

The important thing to remark there is that, if you overwrite a global object, don't "manage the mock from there", keep a reference that is typed as a mock and use that: Don't do `global.mutationObserver.mockClear()`, assign a mock to that property, but use that reference to clear it.

src/components/zen-table-row/test/zen-table-row.spec.tsx
```diff
@@ -1,18 +1,15 @@
...
+
+const originalGetDefaultSlotContent = helpers.getDefaultSlotContent;
```
homer0, line 12: ~~This should be done by mocking `helpers` and using `mockImplementation`~~

You can resolve this comment, but just so you have it present: the thing with the types is really messed up, to the point that you can't mock before importing `jest` from `@jest/globals`... which breaks the flow of mocking.

I just tried and I got so many errors xD.

src/components/zen-table-row/test/zen-table-row.spec.tsx
```diff
@@ -32,51 +29,13 @@ describe('zen-table-row', () => {
...
-    simulateMouse('click', expandableIcon);
-    cleanupTableStructure(table);
-    expect(mainRow.expanded).toBe(true);
-    expect(expandRow.$visible).toBe(true);
-  });
 });
 
-describe('zen-table-row tree functionality', () => {
+describe('zen-table-row inside tree structure', () => {
   let page: SpecPage;
```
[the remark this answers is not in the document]
homer0: Yes, this is missing.

src/components/zen-table-row/test/zen-table-row.spec.tsx
homer0: Yep, this is what I was saying before regarding using a module reference for mocking; this is probably fine for now (until we fix types on the tests), so you'll need to add

```js
// eslint-disable-next-line @typescript-eslint/ban-ts-comment
// @ts-ignore
```

src/components/zen-table-row/test/zen-table-row.spec.tsx
chavlji: Very nice! Thanks for this!

src/components/zen-table-row/test/zen-table-row.spec.tsx
chavlji: Added, but we should solve this globally. I've already added a ticket for that a few weeks ago: https://reciprocitylabs.atlassian.net/browse/PLAT-1800

homer0: I'll try to take of this today.

homer0 (approved):
🤘

PR #319, opened 2021-04-15
Re-enable React bindings deploy

homer0:
Enabling the React bindings should be a 'chore', but we need to actually trigger a release, that's why it's a 'fix'.


f-jauregui (approved):
Let's ship those React bindings!

PR #320, opened 2021-04-16
PLAT-1800: Fix types on the tests

homer0:
[JIRA ticket](https://reciprocitylabs.atlassian.net/browse/PLAT-1800)

#### Description

- The main `tsconfig` now uses the jest types, and on the `cypress` directory there's a new `tsconfig` that extends the one on the root, with the types for cypress.
- I fixed all the type errors and most of the mock implementations on the unit tests; the reason I also changed the mocks it's because most of the type issues were related to that.
- I added a new `typecheck` script to the `package.json` and hooked it on the `pre-push`, like `mf-execution`.
- Yes, the commits are really small in some cases, and it's possible that I could've used one commit for each type of change, but I actually did the changes one component at a time, based on the `tsc` report. I also considered that it could be beneficial to have these small commits as references.

To be clear: I fixed the issues, but there are some cases (like the dropdown) where I know there's a mix of types (`HTMLElement` vs `Element`) and I had to add "unknown conversions" to get it to work; so the TypeScript part, may need a good review in the future.

Regarding the "Jest part", I ended up using spies instead of direct mocks because some cases used some mocked helpers and some real functions, and spies work best in those cases. Normally, you would "mock the entire dependency" and "prepare the implementation".

I also converted some functions to arrow functions; not because it's my preference (although it is :P), but because we had some arrows and some regular functions, and I like to be consistent (when possible).

### DIFF
--- a/src/components/zen-tooltip/test/zen-tooltip.spec.tsx
+++ b/src/components/zen-tooltip/test/zen-tooltip.spec.tsx
@@ -1,18 +1,24 @@
 import { newSpecPage } from '@stencil/core/testing';
-
-const popperMock = {
-  destroy: jest.fn(),
-  state: {
-    placement: 'bottom',
-  },
-};
 import * as popper from '@popperjs/core';
-popper.createPopper = jest.fn(() => popperMock);
-
 import { ZenTooltip } from '../zen-tooltip';
 import { ZenPopover } from '../../zen-popover/zen-popover';
 
 describe('zen-tooltip', () => {
+  beforeEach(() => {
+    jest
+      .spyOn(popper, 'createPopper')
+      .mockClear()
+      .mockImplementation(
+        () =>
+          (({
+            destroy: jest.fn(),
+            state: {
+              placement: 'bottom',
+            },
+          } as unknown) as popper.Instance),
+      );
+  });
+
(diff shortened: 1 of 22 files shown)


src/components/zen-tooltip/test/zen-tooltip.spec.tsx
```diff
@@ -27,7 +33,9 @@ describe('zen-tooltip', () => {
     tooltip.variant = 'error';
     await page.waitForChanges();
     expect(popover.backgroundColor).toEqual('#c22f3d');
-
+    // This is an invalid case; we need to review it, because there's no "empty variant".
```
chavlji, line 36: True. This should be removed.

chavlji (approved):
Awesome clean-up and improvement of tooling!

I love the fact that we finally have a type-check setup before commit! 🚀

src/components/zen-tooltip/test/zen-tooltip.spec.tsx
homer0: @chavlji No worries, I did this before #321 and on that PR, since I removed the variants, I removed this test too. Once I merge this, I'll rebase the other and problem solved.

PR #323, opened 2021-04-20
PLAT-1791: Added disableBeforeDate and disableAfterDate props to zen-date-picker

f-jauregui:
[JIRA ticket](https://reciprocitylabs.atlassian.net/browse/PLAT-1791)

#### Description
As per JIRA ticket

#### ChangeLOG

#### Useful links


homer0 (commented):
Nice 🤘... but I have a couple of comments/questions:

- `disableAfterDate` also disables "the date": If I tell it to disable the dates after `04/22/2021` it will not only disable `23` and on, but also `22`; not sure what the rest thinks, but I think it could be confusing if it says "after" but it also disables that date. Also, the same thing doesn't happen with `disableBeforeDate`, it disables all dates but not "the date".
- I saw that you did the conversion on the `componentDidLoad`; if I'm not wrong, that works for a React based application, but if the attribute were to change without creating a new instance, wouldn't that mean that the value wouldn't be parsed again?
- I think we should probably say somewhere that the `disableBeforeDate` and `disableAfterDate` values should be compatible with the `Date` constructor. Also, not sure if the data type shouldn't be `Date` too, like `value`... otherwise the implementation will need to manage the same component with two different "APIs": JS for `value` and HTML for `disable...`; what do you think?
- When calculating the disabled dates, you store them on `beforeDate` and `afterDate`: I think we should at least add some inline doc about those, because without the `disable` prefix, reading them without context doesn't tell you anything about what they represent. Also, shouldn't they be `private`?

Sorry :P; if you prefer, we can talk about it offline.

src/components/zen-date-picker/zen-date-picker.tsx
```diff
@@ -221,6 +229,14 @@ export class ZenDatePicker {
     } else {
       this.dateChanged(this.value); // set today date
     }
+
+    if (this.disableBeforeDate !== 'null' && this.disableBeforeDate !== null) {
+      this.beforeDate = isValid(new Date(this.disableBeforeDate)) ? new Date(this.disableBeforeDate) : null;
```
chavlji, line 234: refactor: put `new Date(this.disableBeforeDate)` into var
same for `new Date(this.disableAfterDate)`

chavlji (changes requested):
Works as expected! 👍🏻 

One minor thing: I think we should check if date is valid (between before and after) in the `value` and `formattedDate` watcher. If it's not, we should set it to nearest available date.
Eg. if I set `disableAfter` to `04/15/2021` and value to `04/25/2021`, value should be `04/15/2021`

chavlji (commented):
Works better, but I get this error: date is set to `disabledAfter` + 1 day. It should be to `disabledAfter` exactly:

https://user-images.githubusercontent.com/5729421/115710513-b8fb2c80-a372-11eb-8689-dba30b3725b3.mov

f-jauregui:
> Works better, but I get this error: date is set to `disabledAfter` + 1 day. It should be to `disabledAfter` exactly:
> 
>  Screen.Recording.2021-04-22.at.13.55.46.mov

Fixed. @chavlji, please review again! 😄

PR #325, opened 2021-04-20
QUICK: Bugfix - Date picker - empty if formattedDate === today

chavlji:
Bug:
Zen-date-picker: If attr `formatted-date` is set to same date as today, value isn't initialized and calendar is empty:
Screen Shot 2021-04-20 at 10 39 29


f-jauregui (approved):
It works now! 👍

PR #327, opened 2021-04-21
PLAT-1847: Popover is not displayed on the story with filter options 

chavlji:
[JIRA ticket](https://reciprocitylabs.atlassian.net/browse/PLAT-1847)

Popover width is tricky... sometimes we want it to take up whole avail space (dropdown), sometimes we want it custom (date-picker, filter), sometimes we want max 300px or max-content-width (tooltip).


homer0 (commented):
While the solution seems to work as expected on Storybook, I'm not sure removing the control of the `maxWidth` is the best decision here: the implementation should be able to control the dimension of the tooltip.

We can discuss this during the sync.

chavlji:
@homer0 they're still able to do that with a simple:
`<zen-popover style="max-width: 500px" />`
It's the most "natural" and consistent way, since we support this kind of dimensions controlling for all our components.

f-jauregui:
@chavlji we can edit the max-width since the CSS property is part of the component's :host selector, right?

homer0 (approved):

PR #328, opened 2021-04-21
PLAT-1815: Add status tracker component

mihaelpalfi:
[JIRA ticket](https://reciprocitylabs.atlassian.net/browse/PLAT-1815)

#### Description
As per JIRA ticket

image

#### ChangeLOG

#### Useful links


src/components/zen-status-tracker/zen-status-tracker.tsx
```diff
@@ -0,0 +1,63 @@
...
+
+      // disable all the lozenge elements after the selected one
+      if (hasSelected && this.selected != index) lozenge.setAttribute('disabled', '');
+
+      // remove border left after selected lozenge
+      if (this.selected == index - 1) {
+        lozenge.style.borderLeft = '1px solid transparent';
+      }
+    });
+  }
```
homer0, line 43: I have a few suggestions for this method:

- I would separate the loop for when it's archived, that way you don't have to do early exit on every iteration and, since the amount of things it does is just two lines, it will be easier to read.
- Instead of keeping the flag outside the iteration, you could use a reducer and use it as the accumulator.
- When changing the `variant` of the selected, you could use a ternary for the value.
- `===` > `==` :P.
- Try to avoid, if possible, "ternary lines"; they are really hard to differentiate from assignments xD. For a second, I thought you were overwriting the index with the result of `setAttribute` :P.
- When adding the transparent border, you could also check if `index` is truthy, because there's no way the selected one will be `-1`.

Based on those comments, I did this:

```ts
  setChildStyles(): void {
    if (this.archived) {
      this.children.forEach(lozenge => {
        lozenge.setAttribute('variant', 'none');
        lozenge.setAttribute('size', 'lg');
      });

      return;
    }

    this.children.reduce((hasSelected, lozenge, index) => {
      lozenge.setAttribute('size', 'lg');
      const isSelected = this.selected === index;
      if (isSelected) {
        hasSelected = true;
        lozenge.setAttribute(
          'variant',
          index === this.children.length - 1 ? 'green' : 'dark-blue-ghost'
        );
      } else {
        lozenge.setAttribute('variant', 'none');
      }

      if (hasSelected && !isSelected) lozenge.setAttribute('disabled', '');
      if (index && this.selected === index - 1) {
        lozenge.style.borderLeft = '1px solid transparent';
      }

      return hasSelected;
    }, false);
  }
```

To be clear: **It's not "change it to look like that**; the only blocker comments are the `===` and the "ternary line", for the rest, take or ignore what you think it's best, those are just suggestions.

(Too many meetings, I wanted to code something :P).

src/components/zen-status-tracker/zen-status-tracker.tsx
```diff
@@ -0,0 +1,63 @@
...
+
+  componentDidLoad(): void {
+    this.children = Array.from(this.host.children).map(n => n as HTMLZenLozengeElement);
```
homer0, line 46: Not sure if this is the place, but we should probably validate something like

```js
if (this.selected > this.host.children.length - 1) {
  throw new Error("There's no lozenge at the selected index");
}
```

What do you think?

homer0 (commented):
A few comments, mostly syntax-wise, but looks great 🤘!

src/components/zen-status-tracker/zen-status-tracker.tsx
mihaelpalfi: Would not do that. I would like to set like "-1" if a case comes up that nothing should be selected. But would validate if no children were added. Added.

src/components/zen-status-tracker/zen-status-tracker.tsx
mihaelpalfi: Good fed-back! Not so skilled yet. Changed

src/components/zen-status-tracker/zen-status-tracker.stories.mdx
```diff
@@ -0,0 +1,99 @@
+import { Meta, Story, Canvas, ArgsTable } from '@storybook/addon-docs/blocks';
+import { html } from 'lit-html';
+import { getArgTypes, getDefaultArgs, getComponentData, spreadArgs } from '#storybook/helpers/argTypes';
+
+import data from '../../../stencilDocs.json';
+const compData = data.components.find(n => n.tag === 'zen-status-tracker');
+const argTypes = getArgTypes(compData);
+
+<Meta title="Navigation/Status Tracker" component="zen-status-tracker" argTypes={argTypes} />
```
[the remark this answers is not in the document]
mihaelpalfi: All other components titles are in uppercase. Its the progress tracker issue. :)

src/components/zen-status-tracker/zen-status-tracker.tsx
```diff
@@ -0,0 +1,65 @@
...
+
+      if (hasSelected && !isSelected) lozenge.setAttribute('disabled', '');
+      if (index && this.selected === index - 1) lozenge.style.borderLeft = '1px solid transparent';
+
+      return hasSelected;
+    }, false);
+  }
+
+  componentDidLoad(): void {
+    if (this.host.children.length == 0) throw new Error("There's no lozenge children");
```
chavlji, line 47: This function should run in mutation observer. Currently it will only run on load and if lozenges are added/deleted during this component's lifespan, state won't be updated.

src/components/zen-status-tracker/zen-status-tracker.tsx
```diff
@@ -0,0 +1,65 @@
...
+  tag: 'zen-status-tracker',
+  styleUrl: 'zen-status-tracker.scss',
+  shadow: true,
+})
+export class ZenStatusTracker {
+  children: HTMLZenLozengeElement[];
+  @Element() host: HTMLZenStatusTrackerElement;
+
+  /** Selected index */
+  @Prop() readonly selected: number = 0;
```
chavlji, line 14: Why do we need this prop? It would be better to put this prop to lozenga component instead.
User could just add lozenge with the correct attributes.
Imo this component should only take care of correctly showing dividers. If we wouldn't have those, we wouldn't even need this component. Container with display:flex row would be enough...

src/components/zen-status-tracker/zen-status-tracker.tsx
```diff
@@ -0,0 +1,65 @@
...
+
+  /** Status archived */
+  @Prop() readonly archived: boolean = false;
```
chavlji, line 17: Same as selected. Put this into lozenga component instead.

chavlji (changes requested):
Works ok, but I'd set things a bit differently.

We're mixing 2 approaches: semantic web and reactive framework component approach. Since we have a semantic approach here, we should let things be set on children and loosen the control from the parent.

src/components/zen-status-tracker/zen-status-tracker.tsx
homer0: I kind of disagree... but I initially thought of the same thing :P.

First, I thought it didn't make a lot of sense  to make the component reusable, as you said, the dividers it's the only "feature" that provides and the rest could be solved with `flex-direction: row`... but then, it was confirmed that there will be multiple cases for this, which can mean multiple apps, so it was bound to have repetition.

Now, the product team designed this component as a composition of specific states (color), variants (size) and order (`archived`), and thinking on multiple implementations, it makes sense for it to already have its logic defined in a single place that can easily maintained.

If we were to only take care of the dividers, we would be splitting the component responsibilities into multiple places: The lozenge for the archived, this component for the dividers and the implementation for the states; that's too many places to maintain a single functionality, even one as small as this one.

We can discuss this further on the sync.

src/components/zen-status-tracker/zen-status-tracker.tsx
homer0: Same argument as above :P

src/components/zen-status-tracker/zen-status-tracker.tsx
chavlji: @homer0  I also gave this a further thought yesterday afterward :) My base argument would be just to relieve this component of so many responsibilities. But I totally agree with your reusability argument.

If it will stay like this we need to fix this, to work as expected:
1. `selected` prop should be item `id` and not just index (if user mutates children, the incorrect item might be highlighted)
2. `selected` and `archived` classes/styles should be recalculated in mutation observer (basically the same thing as in my first comment)

homer0: Awesome :D.

1. Up to you guys, I have no strong preference over that; the only thing is that we'll also need to validate that all children have an `id` attribute.
2. Agree 👍

mihaelpalfi: Done

src/components/zen-status-tracker/zen-status-tracker.tsx
mihaelpalfi: Done

src/components/zen-status-tracker/zen-status-tracker.tsx
```diff
@@ -0,0 +1,117 @@
...
+  }
+
+  isValid(): boolean {
+    if (!this.hasChildren()) {
+      console.error('zen-status-tracker: There is no `zen-lozenge` elements!', this.host);
+      return false;
+    }
+
+    if (!this.hasIdsSet()) {
+      console.error('zen-status-tracker: Not all of `zen-lozenge`elements have id defined!!', this.host);
```
homer0, line 40:
```suggestion
      console.error('zen-status-tracker: Not all of `zen-lozenge` elements have id defined!!', this.host);
```

src/components/zen-status-tracker/zen-status-tracker.tsx
```diff
@@ -0,0 +1,117 @@
...
+  hasIdsSet(): boolean {
+    return this.getChildren().every(item => item.hasAttribute('id'));
+  }
+
+  hasChildren(): boolean {
+    return this.getChildren().length > 0;
+  }
+
+  isValid(): boolean {
+    if (!this.hasChildren()) {
```
homer0, line 34: I agree with what we discussed on the daily about having the `has[x]` and the `get[x]` in separated methods, in this case, I don't see a point on having a method for only one execution context: `hasChildren` is only being used in this method, but then we are using another method (`hasIdsSet`) that also needs the children.

I would recommend doing something like you did on `setLozengeProperties`, capture the children and then use them multiple times:

```js
const children = this.getChildren();

if (!children.length) {
  ...
}

if (!children.every(item => item.hasAttribute('id')) {
  ...
}

...
```

homer0 (approved):
🤘!!! (Small syntax-releated comment)

homer0:
@mihaelpalfi If you are ready to merge and the `coverage` job keeps failing for the same thing, you can ignore it... that's on coveralls.

src/components/zen-status-tracker/zen-status-tracker.tsx
mihaelpalfi: Fixed

src/components/zen-status-tracker/zen-status-tracker.tsx
mihaelpalfi: Changed.

PR #329, opened 2021-04-22
QUICK-FIX(zen-tabs): prevent selected tab jumping

chavlji:
Eliminate ugly tabs jumping when selecting a tab. Uses `text-shadow` instead `font-weight`.
Difference is imo very hard to notice even if you compare it side by side:
Screenshot 2021-04-22 at 11 20 28

### DIFF
--- a/src/components/zen-tab/zen-tab.scss
+++ b/src/components/zen-tab/zen-tab.scss
@@ -9,6 +9,7 @@
   border-bottom: 4px solid transparent;
   color: $color-gray-400;
   font-size: $text-md;
+  letter-spacing: 0.5px;
 
   ::slotted(*) {
     color: $color-gray-400;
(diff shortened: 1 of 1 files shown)


mihaelpalfi (approved):

f-jauregui (changes requested):
I'm not sure about this, changes in the typography silhouette are noticeable. There's no way to prevent the jump without a hack?

chavlji:
@f-jauregui Not an easy one.

We would need to make original content transparent. Then we should add content duplicate absolutely positioned in place of original content. And duplicated content should have `font-weight: bold`.

But text is not just a text. It's a slot. So we'd need to copy a slot and if user would update the original slot content... 🐇 🕳️

homer0:
Moving this `WIP` until Jan gets back.

chavlji:
I'm closing this one since the solution is not ideal. The best solution would be to rethink the design and to not use bold text in such cases (unselected - normal, selected - bold).

PR #330, opened 2021-04-23
PLAT-1978: Change events name to to zen prefix ones

mihaelpalfi:
All events, except for `click`, now use the `zen` prefix. For example `onChange` is now `onZenChange`.

[JIRA ticket](https://reciprocitylabs.atlassian.net/browse/PLAT-1978)

#### Description

- Changed all onChange to zenChange events.
- Additionally changed most of other events like onBlur, onFocus, onInput. 

Issues that still remains and have to be looked at:

- [x] Check if an event was missed for a component. There are to much of them :) 
- [x] Some components trigger events 2 times. (zen-calendar, ...)
- [x] Some events are not being triggered or logged. (zen-panel, ...)
- [x] Events that are triggered from a child component, should probably have the element which triggered it included.
- Fixed and tried to improve the tests. But still the coverage for components decreased because of the event changes. 

---

Changes by homer0:

- I added checkboxes to Mihael comments because I went through all the components, I checked the events the implemented, the events the listened for and if Storybook was logging them. Thanks to this I was able to fix a couple of things that were already on `main`, like missing mutation on the `zen-tabs` and missing `logEvents` on the `zen-popover` stories.
- The date picker was triggering twice because it was also triggering the `zen-input` events; I stopped the propagation of the those events.
- I changed the target of the branch to a feature branch because I want to fix a couple of types so the implementation can properly do `event.target` without the need of casting to the element they know triggered the event.

### DIFF
--- a/src/components.d.ts
+++ b/src/components.d.ts
@@ -699,6 +699,10 @@ export namespace Components {
          */
         "disabled": boolean;
         /**
+          * Method triggers a custom event "zenChange"
+         */
+        "dispatchChangeEvent": () => Promise<void>;
+        /**
           * Group id to which this radio belongs
          */
         "group": string;
@@ -1771,6 +1775,10 @@ declare namespace LocalJSX {
          */
         "name"?: string;
         /**
+          * Checkbox change event
+         */
+        "onZenChange"?: (event: CustomEvent<void>) => void;
+        /**
           * Shows a red asterisk after label.
          */
         "required"?: boolean;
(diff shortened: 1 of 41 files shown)


homer0:
- This should affect ALL events but click; we should be consistent and avoid forcing the people implementing it (probably us) to validate with the documentation which events have `Zen` and which don't, because this affects all native events, like `input`.
- You forgot the double colon after `BREAKING CHANGE` :P.

homer0:
@f-jauregui sorry to bother you again with this... can you update the snapshots :P?

f-jauregui (approved):
👍

PR #335, opened 2021-04-26
PLAT-1994: Create zen-sidebar-nav-subitem component

chavlji:
[JIRA ticket](https://reciprocitylabs.atlassian.net/browse/PLAT-1994)
Based on https://github.com/reciprocity/zen-ui/pull/334

### DIFF
--- a/src/components/zen-sidebar-nav-item/zen-sidebar-nav-item.tsx
+++ b/src/components/zen-sidebar-nav-item/zen-sidebar-nav-item.tsx
@@ -1,4 +1,4 @@
-import { Component, Host, h, Prop, Event, EventEmitter, Watch, Element } from '@stencil/core';
+import { Component, Host, h, Prop, Event, EventEmitter, Watch, Element, State } from '@stencil/core';
 import { applyPrefix, getSlotElement } from '../helpers/helpers';
 import { faChevronDown } from '@fortawesome/pro-regular-svg-icons';
 
@@ -9,6 +9,8 @@ import { faChevronDown } from '@fortawesome/pro-regular-svg-icons';
 })
 export class ZenSidebarNavItem {
   @Element() host: HTMLZenSidebarNavItemElement;
+
+  @State() hasSubitems = false;
 
   /** Render item as selected */
   @Prop({ reflect: true }) readonly selected: boolean = false;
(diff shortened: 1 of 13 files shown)


homer0 (approved):
- Remove the `hover` functionality

:shipit:

chavlji:
@homer0 @f-jauregui are you sure about removing expand on hover? Nina told me on tueseday it's supposed to work like this.

f-jauregui:
@chavlji I'll let this PR open in case that we want to preserve that functionality. In that case, I'll put that back again.
Let's wait for @homer0's review. But as far as I talked to him yesterday this is not necessary for alpha.

homer0:
@chavlji go enjoy your holidays xD! Regarding the collapse functionality:

- Alpha is going out with the toggle disabled: we still don't have the user preferences service, we are not persisting user settings, so instead of resetting the state on each refresh, we'll remove the whole functionality. We validated the decision with Shelly.
- The toggle button works like a charm, and since you already coded it, we'll just hide it with a prop... but the hover functionality looks a little bit hacky (:P, sorry) and the rule of thumb was "if we have to sit down and check what is the best approach for a functionality that we don't want to deliver.... yep, it's not happening".

src/components/zen-sidebar-nav-item/zen-sidebar-nav-item.tsx
```diff
@@ -1,27 +1,75 @@
...
+    const others = this.getItems().filter(item => item.selected && item !== newlySelected);
+    others.forEach(item => {
+      item.selected = false;
+    });
+    this.selected = true;
+  }
+
+  componentDidLoad(): void {
+    // TODO: Do this in mutation observer!
+    this.hasSubitems = !!this.getItems().length;
```
homer0, line 58:
```ts
startChildObserver(): void {
  this.childObserver = new MutationObserver(() => {
    this.hasSubitems = this.getItems().length > 0;
  });

  this.childObserver.observe(this.host, {
    childList: true,
    attributes: true,
    subtree: true,
  });
}
```

homer0 (approved):
🤘

chavlji:
😄 👍🏻

PR #339, opened 2021-05-03
PLAT-2020: Add tabs improvements

mihaelpalfi:
[JIRA ticket](https://reciprocitylabs.atlassian.net/browse/PLAT-2020)
AND
[JIRA ticket](https://reciprocitylabs.atlassian.net/browse/PLAT-2034)

#### Description
As per JIRA ticket

image

#### ChangeLOG

#### Useful links


src/components/zen-tab/test/zen-tab.spec.tsx
```diff
@@ -10,4 +11,36 @@ describe('zen-tab', () => {
...
+    expect(page.root.shadowRoot.querySelector('.badge')).toBeTruthy();
+  });
+
+  it('should be disabled', async () => {
+    const page = await newSpecPage({
+      components: [ZenTab],
+      html: `<zen-tab disabled>Tab 1</zen-tab>`,
+    });
+
+    expect(page.root.hasAttribute('disabled')).toBeTruthy();
```
homer0, line 30:
```suggestion
    expect(page.root.hasAttribute('disabled')).toBe(true);
```

This returns boolean, so we can compare it to `true` directly.

src/components/zen-tab/test/zen-tab.spec.tsx
```diff
@@ -10,4 +11,36 @@ describe('zen-tab', () => {
...
+      components: [ZenTab],
+      html: `<zen-tab>Tab 1</zen-tab>`,
+    });
+
+    const clickSpy = jest.fn();
+    page.root.addEventListener('click', clickSpy);
+    simulateMouse('click', page.root);
+    await page.waitForChanges();
+
+    expect(clickSpy).toHaveBeenCalled();
```
homer0, line 44: Not a blocker, but I would suggest you to use `toHaveBeenCalledTimes` over `toHaveBeenCalled` as it prevents undesired side effects (like the double `focus` we had on the date picker).

```suggestion
    expect(clickSpy).toHaveBeenCalledTimes(1);
```

src/components/zen-tab/zen-tab.scss
```diff
@@ -1,7 +1,7 @@
 @import '../scss/component-helpers.scss';
 
 :host {
-  display: block;
+  display: flex;
```
homer0, line 4: I always forget this (sorry!)... but didn't we say that we should always be `block` and the implementation should be the one defining "flex rules"? or was it the other way around? @chavlji ?

src/components/zen-tab/zen-tab.stories.mdx
```diff
@@ -26,6 +26,7 @@ A tab fragment component to be used in `zen-tabs`.
     name="default"
     args={{
       ...getDefaultArgs(argTypes),
+      badge: null,
```
homer0, line 29: Why do we have to add this only for `badge`? it also has a default value.

homer0 (commented):
🤘 - Super small comments and a question :P, but looks great :D.

homer0:
One more thing (:P), the scope for the `feat` commit should be `zen-tab`, not `zen-tabs`.

src/components/zen-tab/zen-tab.stories.mdx
mihaelpalfi: Storybook controls always send a string value. So it will send 'null' and the badge will still be displayed. Probably this is not the issue when using it in the app. Correct type will be used. Right?

src/components/zen-tab/zen-tab.scss
mihaelpalfi: Hmm dont remember. If i dont set it to flex i have to wrap the content in another tag.

src/components/zen-tab/test/zen-tab.spec.tsx
mihaelpalfi: Done

src/components/zen-tab/test/zen-tab.spec.tsx
mihaelpalfi: Done

src/components/zen-tab/zen-tab.scss
homer0: Let's wait for @chavlji to confirm it, but since the use case of the `zen-tab` is inside of a `zen-tabs`, maybe we don't have to be that much strict about this.

src/components/zen-tab/zen-tab.stories.mdx
homer0: Ahh, makes sense; in order to be able to send it as `null`, you have to interact with it via JS, otherwise it will always be `''`; and I agree with you, no need to "hotfix it just for Storybook".

(You can resolve the comment :D).

src/components/zen-tab/zen-tab.scss
chavlji: @homer0 I remember we were talking about this, but I can't recall the exact case and I'd say it very much depends.
I'd say flex is ok here since the content of a tab is expected to be layouted in a row.

chavlji (changes requested):
Please make `Badge` a standalone component. It may be used on various occasions (inside buttons, icons,...).
The component shouldn't be responsible for its content. That's why we have slots.

homer0:
@chavlji we talked about this on the daily yesterday, the badge is on the plans, but it won't be a component yet; we agreed that when the component gets defined, we'll consume it inside the tabs. It's not like we are going to "do a lot of rework", it's a `div`.

homer0 (approved):

chavlji:
@homer0 I don't agree with this approach. This way we'll constantly be presenting breaking changes (by removing props).

I remember we've agreed only on creating such things in the React and porting it to zen-ui later if we see the use case. This way it would be ok, because we're not introducing breaking changes.

Creating a standalone component takes some 20% more time than solving the problem with the prop. Communicating breaking change and implementing changes for everyone involved takes a lot more than creating such a small component in the first place.

Zen-ui is a core library and we should keep it clean.

homer0:
@chavlji it wouldn't be a breaking change because we wouldn't make it a slot: a tab is not a container, and it doesn't have multiple contexts (like a button). If we were to add an extra component and a slot for showing just a number next to the title we would increase not only the maintainability complexity, but also the one on the implementation.

Regardless of breaking changes or not, IMO, it's an unnecessary abstraction; when the component for the badge gets created, we should implement it, but still inside the tab component.

We can discuss this on the daily with the rest of the team.

mihaelpalfi:
@homer0 @chavlji i can add the component no problem at all. Half an hour of work. But that you put high priority on the prs.

chavlji:
@homer0 it's already a container. It already accepts the default slot. That's how you pass the label currently.

And in fact, this is the only correct way. If someone wants to add an icon into the tab or whatever, he's free to do so... This is how you make component flexible. Otherwise, we'd have to support each and every tab use case (without icon, with icon, with icon and badge, with the avatar,...). Whichever the designers will come up with. We should let apps the freedom to combine basic components however they like. Not just how we've designed them to be combined.

Plus imo it's less complex because components are smaller and they have less responsibility.

And as @mihaelpalfi  mentioned it doesn't take much more time to do things the right way.

homer0:
@chavlji it's a container in the same way a `label` is a container, is not like a `button` that can have any kind of content and can be used on any kind of context. Flexibility just because we can is not a good reason to do so, we'll end up maintaining more cases that needed just because we want to accommodate any possible scenario.

Now, let's say we go with a specific component:

- We'll either create it without specs (since this doesn't exist on Product side), or we'll be blocked by the definition of the specs.
- If we go with the first option, there you have the possibility of breaking changes.
- We add a slot on the `zen-tab`, we need to support multiple positions, or are we having a leading slot and trailing slot? Does it make sense to only put the slot on the right?

I don't think all of that is justified for a label on a tab; if you still strongly believe this should be implemented as a slot, we can discuss it on the daily and create a TD: we talked about this yesterday, and we already have it ready.

chavlji:
@homer0 We already have a slot. The slot can already accept anything. You can't force slot to only accept text nodes. And in the end, why would you even want that? We already have an element inside tab, that is not a text node - a badge.

We should definitely discuss this on a daily. I'm really sorry I've misheard it yesterday 😢 
If we're not going to go with the semantic web paradigm it's a total shift of the zen-ui. This is a very big decission.

chavlji (approved):
We'll convert badge to component in a separate task later.

PR #342, opened 2021-05-05
PLAT-2048: Add sidebar loading state component

chavlji:
[JIRA ticket](https://reciprocitylabs.atlassian.net/browse/PLAT-2048)


mihaelpalfi (approved):
Looks fantastic! Just a question. Why is this a separate component and not a part of sidebar nav?

homer0:
🤘! But remove the skeleton for the program picker, that's on the implementation side (not all the menus will have it).

chavlji:
@homer0 shall I leave it as a prop `showProgramSelector`?

@mihaelpalfi We'd clutter the original component. Since it can be easily separated and it doesn't overlap very much, imo it's better to have 2 separate components in this case.

homer0:
@chavlji I would remove it completely; we are assuming implementation context and this will be used on all apps; if for some reason, the implementation changes the picker, or they add something else, it would invalidate the skeleton here, so if whatever goes above/below the menu is part of the implementation, the implementation should be in charge of its skeleton too.

chavlji:
@homer0 I've removed it from the component and added story for how to create program selector.

PR #343, opened 2021-05-06
PLAT-2101: Bug: Status tracker observer loops infinitely

chavlji:
- setAttribute always triggers observer callback.
- added a check to set it only if it actually differs.

[JIRA ticket](https://reciprocitylabs.atlassian.net/browse/PLAT-2101)

Takeaway:
Always test observers by mutating children in manually from the dev tools.

To discuss:
Could we add unit tests to test such behaviour?


homer0 (approved):
🤘

mihaelpalfi (approved):
Thank you for fixing this!

PR #344, opened 2021-05-10
PLAT-2134: ZenText ellipsis doesn't work and ZenTableCell has to have a style property

chavlji:
[JIRA ticket](https://reciprocitylabs.atlassian.net/browse/PLAT-2134)

Added props and new story with different truncation options:
Screenshot 2021-05-10 at 10 11 04


mihaelpalfi (approved):
Nice. Tested and it works. Just update the visual tests.

PR #346, opened 2021-05-13
PLAT-2179: [zen-ui] Checkbox indeterminate state is broken

chavlji:
Instead of embedding svg, we should just use zen-icon as with all
other components.

[JIRA ticket](https://reciprocitylabs.atlassian.net/browse/PLAT-2179)


mihaelpalfi (approved):
LGTM

PR #348, opened 2021-05-17
PLAT-2242: Zen-UI: Add disabled prop to zen-sidebar-nav-item

chavlji:
[JIRA ticket](https://reciprocitylabs.atlassian.net/browse/PLAT-2242)

### DIFF
--- a/src/components/zen-sidebar-nav-item/zen-sidebar-nav-item.tsx
+++ b/src/components/zen-sidebar-nav-item/zen-sidebar-nav-item.tsx
@@ -20,12 +20,16 @@ export class ZenSidebarNavItem {
   /** Item was selected */
   @Event() zenSelect: EventEmitter<void>;
 
+  /** Item disabled */
+  @Prop({ reflect: true }) readonly disabled: boolean = false;
+
   @Watch('selected')
   selectedChanged(selected: boolean): void {
     if (!selected) {
       this.deselectAllChildren();
       return;
     }
+    if (this.disabled) return;
     this.zenSelect.emit();
   }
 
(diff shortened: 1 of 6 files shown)


src/components/zen-sidebar-nav-item/zen-sidebar-nav-item.tsx
```diff
@@ -20,16 +20,16 @@ export class ZenSidebarNavItem {
...
+  @Prop({ reflect: true }) readonly disabled: boolean = false;
+
   @Watch('selected')
   selectedChanged(selected: boolean): void {
     if (!selected) {
       this.deselectAllChildren();
       return;
     }
-    const subitems = this.getItems();
-    if (subitems.length && !subitems.some(n => n.selected)) {
```
homer0, line 30: While this makes sense (I think the ticket for improvements mentions something related), how do we "open" an item?

chavlji: It still works fine. It just remove autoselect of the first child element (which is in our case Controls item which will be disabled):

https://user-images.githubusercontent.com/5729421/118478775-8e518900-b710-11eb-8889-a6bcb71d40c0.mov

I know it's still not perfect, but fixing it exactly as it should be, would take me much more time and I won't be ready till due date. We do have a [separate ticket to fix it as it should be](https://reciprocitylabs.atlassian.net/browse/PLAT-2103).

homer0 (approved):

src/components/zen-sidebar-nav-item/zen-sidebar-nav-item.tsx
homer0: Sorry! I thought that it would only open if one of its children was selected.

PR #352, opened 2021-05-24
PLAT-1904: Add system tooltip variant and arrow

mihaelpalfi:
[JIRA ticket](https://reciprocitylabs.atlassian.net/browse/PLAT-1904)

#### Description
As per JIRA ticket

@Donaab Would you please be soo good and check why the vrts are failing now,  i added the variant story.

#### ChangeLOG

#### Useful links


src/components/zen-tooltip/zen-tooltip.tsx
```diff
@@ -56,7 +60,19 @@ export class ZenTooltip {
...
+      case 'system':
+        this.backgroundColor = '#1e272c';
+        break;
+      default:
+        this.backgroundColor = '#FFFFFF';
+        break;
+    }
+  }
   componentDidLoad(): void {
+    this.variantChanged(this.variant);
```
homer0, line 75: Not sure if the watch is due to a convention you were using, but in this case, you don't need it: changing the attribute will already trigger a re-render, so you can have a map of the backgrounds and get the right now, with fallback to `default`, on `render`:

```ts
private backgroundByVariant: Record<TooltipVariant, string> = {
  system: '#1E272C',
  default: '#FFFFFF',
};
// The, on render:
const backgroundColor = this.backgroundByVariant[this.variant] || this.backgroundByVariant.default;
```

I just tested it and it seems to work :D.

The reason I added the comment is not because of the syntax, but since `render` gets triggered anyways, whether you watch the attribute or not, I would assume that one less watcher would be better for performance :P.

mihaelpalfi: Yes it was a convention. But i also dont think this watchers are needed. Changed and added also additional parameter for arrow.

homer0 (approved):
🤘

src/components/scss/vars/shadows.scss
```diff
@@ -3,5 +3,6 @@
 $shadow-focus: 0px 0px 0px 2px $color-blue-300;
 $shadow-invalid: 0px 0px 0px 2px $color-red-300;
 $shadow-drop-200: 0px 1px 2px rgba($color-gray-700, 0.25), 0px 2px 8px rgba($color-gray-700, 0.2);
+$shadow-drop-300: 0px 2px 5px 1px rgba($color-blue-800, 0.15);
```
chavlji, line 6: nit: We could mention it's blue color in the name? Like $shadow-drop-blue-300

chavlji, line 6: nit: We could mention it's blue color in the name? Like $shadow-drop-blue-300

src/components/zen-tooltip/test/zen-tooltip.spec.tsx
```diff
@@ -85,4 +85,20 @@ describe('zen-tooltip', () => {
     expect(heading).toBeTruthy();
     expect(link).not.toBeTruthy();
   });
+
+  it('should render variant system', async () => {
```
chavlji, line 89: No need for this test. It only tests Stencil's mechanism for converting props to attrs.
Test if `backgroundColor` and `showArrow` is set correctly for variant `system` instead.

src/components/zen-tooltip/zen-tooltip.stories.mdx
```diff
@@ -25,6 +25,21 @@ export const Template = args => {
   `;
 };
 
+export const TemplateVariants = () => {
+  return html/*html*/ `
+    <sb-zen-space nowrap horizontal-align="space-evenly" vertical-align="end" spacing="sm">
+      <div>
+          <sb-zen-button label="Default" variant="blue-ghost"></sb-zen-button>
+          <sb-zen-tooltip position="top" label="Lorem Ipsum is simply dummy text of the printing and typesetting industry. Lorem Ipsum has been the industry's standard dummy text ever since the 1500s. Lorem Ipsum is simply dummy text of the printing and typesetting industry."></sb-zen-tooltip>
```
chavlji, line 33: nit: Please helper function `loremIpsum` from storybook utils.js to make it more readable.

chavlji (approved):
Nice! Just a few nits if you have time...

src/components/scss/vars/shadows.scss
mihaelpalfi: removed. Not part of this ticket.

src/components/zen-tooltip/zen-tooltip.stories.mdx
mihaelpalfi: Heh. Didnt know this existed. Added.

src/components/zen-tooltip/test/zen-tooltip.spec.tsx
mihaelpalfi: Done

cypress/integration/visual/tooltip.e2e.js
```diff
@@ -1,8 +1,9 @@
 /// <reference types="cypress" />
 
 describe('Tooltip visual tests', { scrollBehavior: 'center' }, () => {
-  const pageId = 'notifications-tooltip--slot';
+  const pageId = 'notifications-tooltip--variant';
   const story = [
+    'story--notifications-tooltip--variant',
```
Donaab, line 6: @mihaelpalfi Just skip the test for this story and I will fix it.

Donaab (changes requested):

cypress/integration/visual/tooltip.e2e.js
mihaelpalfi: Done

Donaab (approved):

PR #353, opened 2021-05-24
PLAT-2264 - Fix jumping behavior the table

chavlji:
[JIRA ticket](https://reciprocitylabs.atlassian.net/browse/PLAT-2264)

### DIFF
--- a/src/components/zen-table-cell/zen-table-cell.tsx
+++ b/src/components/zen-table-cell/zen-table-cell.tsx
@@ -63,7 +63,7 @@ export class ZenTableCell {
     this.parentRow().selected = (event.target as HTMLZenCheckboxElement).checked;
   }
 
-  transscriptParentRowProps(): void {
+  loadParentRowProps(): void {
     const parentRow = this.parentRow();
     if (!parentRow) return;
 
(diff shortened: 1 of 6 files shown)


mihaelpalfi (approved):
Looks good. Remember that if you specifically dont set the property type it will not translate the properties correctly to react. But for this $ properties its not needed right!
NIT: Still see some jumping when switching tabs. But this is probably another issue.

src/components/zen-table-cell/zen-table-cell.tsx
```diff
@@ -63,6 +63,27 @@ export class ZenTableCell {
     this.parentRow().selected = (event.target as HTMLZenCheckboxElement).checked;
   }
 
+  transscriptParentRowProps(): void {
```
homer0, line 66: Being reaaaally picky here, sorry: transcript has only one `s`, and in any case, what do you think about `inherit` or `load` :P?

homer0 (approved):
@mihaelpalfi Don't worry about the props, the issue is when they have the `readonly`, but he removed it.

@chavlji awesome work 🤘 - just one small comment about naming :P (sorry!)

src/components/zen-table-cell/zen-table-cell.tsx
chavlji: :) Np, fixed.

PR #354, opened 2021-05-24
PLAT-2309: Fix display of initials if username contains on…

mihaelpalfi:
…ly space

[JIRA ticket](https://reciprocitylabs.atlassian.net/browse/PLAT-2309)

#### Description
As per JIRA ticket

Could reproduce on dev. But dont have single spa set locally yet to properly test. There are possibly two issues : 

1. ~~Issue is because the email for a newly invited user is not set. The avatar component automatically uses email if no username for initials. I dont see the email value it in the console data. Has to be fixed on details page. ~~
Email is set so it must be reason 2.   

2. Also another issue is because it seems it even that we check if(username) and its empty it goes into the condition. String with space is set for the username data. Has to be fixed on details page.

user-name=" "
Will add additional check for space in username and email to avatar icon component. Should fix the issue!
#### ChangeLOG

#### Useful links

### DIFF
--- a/src/components/zen-avatar-icon/zen-avatar-icon.tsx
+++ b/src/components/zen-avatar-icon/zen-avatar-icon.tsx
@@ -36,7 +36,7 @@ export class ZenAvatarIcon {
     if (this.initials) return this.initials;
 
     let initials = '';
-    if (this.userName) {
+    if (this.userName && this.userName.trim() != '') {
       if (/\s/.test(this.userName)) {
         // Get initials from name and surname
         initials = this.userName
@@ -48,8 +48,10 @@ export class ZenAvatarIcon {
         // Get initials oly from name
         initials = this.userName.substring(0, 2).toUpperCase();
       }
-    } else {
+    } else if (this.email && this.email.trim() != '') {
       initials = this.email.substring(0, 2).toUpperCase();
+    } else {
+      console.error('zen-avatar-icon : Username or email has to have a value!');
     }
     return initials;
   }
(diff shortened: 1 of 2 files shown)


src/components/zen-avatar-icon/zen-avatar-icon.tsx
```diff
@@ -36,7 +36,7 @@ export class ZenAvatarIcon {
     if (this.initials) return this.initials;
 
     let initials = '';
-    if (this.userName) {
+    if (this.userName && this.userName.trim() != '') {
```
homer0, line 39:
```suggestion
    if (this.userName.trim()) {
```

Always `!==`... but in this case, you can use it as a truthy value.

src/components/zen-avatar-icon/zen-avatar-icon.tsx
```diff
@@ -48,8 +48,10 @@ export class ZenAvatarIcon {
         // Get initials oly from name
         initials = this.userName.substring(0, 2).toUpperCase();
       }
-    } else {
+    } else if (this.email && this.email.trim() != '') {
       initials = this.email.substring(0, 2).toUpperCase();
```
homer0, line 52:
```suggestion
    } else if (this.email.trim()) {
      initials = this.email.trim().substring(0, 2).toUpperCase();
```

Yep, it's repeating the `.trim()`, but if you validate the string with `trim` and then do the `substring` without using `trim`, you could end up with a space.

homer0 (approved):
Tiny details 🤘

src/components/zen-avatar-icon/zen-avatar-icon.tsx
mihaelpalfi: Good catch. But not because of the shorter code.  But because considering the imageUrl as well. Done

src/components/zen-avatar-icon/zen-avatar-icon.tsx
mihaelpalfi: Done

homer0 (approved):
🤘

PR #356, opened 2021-05-25
PLAT-2237: When on slow network on refresh we get an error on the table component

chavlji:
[JIRA ticket](https://reciprocitylabs.atlassian.net/browse/PLAT-2237)


mihaelpalfi (approved):
👍

homer0 (approved):
🤘 If it works, awesome, but out of curiosity, how would be the flow in which a prop is updated, but the parent element is not present?

PR #359, opened 2021-05-27
PLAT-2401: Improve bulk-select events on zen-table

chavlji:
[JIRA ticket](https://reciprocitylabs.atlassian.net/browse/PLAT-2401)

### DIFF
--- a/src/components/zen-table-cell/zen-table-cell.tsx
+++ b/src/components/zen-table-cell/zen-table-cell.tsx
@@ -46,8 +46,12 @@ export class ZenTableCell {
   /** Row is placed right after header (auto calculated) */
   @Prop({ reflect: true, attribute: 'after-header' }) readonly $afterHeader: boolean = false;
 
+  isFirstCell(): boolean {
+    return !this.host.previousElementSibling;
+  }
+
   showWidgets(): boolean {
-    const isFirstCell = !this.host.previousElementSibling;
+    const isFirstCell = this.isFirstCell();
     return isFirstCell && (this.$selectable || this.$expandable);
   }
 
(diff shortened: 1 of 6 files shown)


src/components/zen-table-cell/zen-table-cell.tsx
```diff
@@ -59,10 +63,19 @@ export class ZenTableCell {
...
   }
 
+  onCheckboxClick(event: CustomEvent): void {
+    const payload = {
+      detail: { checked: (event.target as HTMLZenCheckboxElement).checked },
+      bubbles: true,
+      composed: true,
+    };
+    this.host.dispatchEvent(new CustomEvent('cellcheckboxclick', payload));
+  }
```
homer0, line 77: - Why `dispatchEvent` and not Stencil's event emitter?
- Shouldn't the event be prefixed?

src/components/zen-table/zen-table.tsx
```diff
@@ -17,6 +17,18 @@ export class ZenTable {
   /** Table cleanup in progress */
   @Prop({ attribute: 'updating' }) readonly $updating: boolean = false;
 
+  /** Returns array with ids of selected rows (ordered as they appear in the dom) */
+  @Method()
+  async getSelectedRows(): Promise<string[]> {
+    return this.getAllRows()
+      .filter(row => row.selected)
```
homer0, line 24: This is where we should get rid of rows without IDs, right?

```suggestion
      .filter(row => rowId && row.selected)
```

src/components/zen-table-cell/zen-table-cell.tsx
chavlji: @homer0  That's intentional, because I don't want to present this event on the cell but on the row instead. And since it's only internal thing it isn't prefixed.

homer0 (approved):
🤘

PR #366, opened 2021-06-04
PLAT-2471: Tooltip improvements change brake mf-details components usage

chavlji:
[JIRA ticket](https://reciprocitylabs.atlassian.net/browse/PLAT-2471)

### Cause of the bug
[Chromium bug](https://stackoverflow.com/questions/2637058/positions-fixed-doesnt-work-when-using-webkit-transform) (unresolved for 11 years!) disables `position: fixed` to any descendant of a parent with css prop `transform` set. The position is still reported as `fixed`, but it behaves and is positioned as it would be `absolute`.

### Solution
In the particular case, we've left with `transform: translate(0)` set on the drawer after opening transition finished. So solution in all such cases is to animate from eg. `transform: translate(-100%, 0)` into `transform: none`! The animation will still happen, but the ending state won't have any transform applied. I've also removed `translate(0)` everywhere else.

### Solution 2
We can't know in what exact circumstances the popover will be used. That's why I'm checking if any parent has `transform` set right **before popover opening**. In case it does, I'm switching off poppers `strategy: fixed` and showing the console warning:
Screenshot 2021-06-04 at 11 18 13

In such cases, popover will be shown in the correct place, but it might trigger scrollbars of the container or it might be cut of (in case of parent's `overflow: hidden`):

**position: absolute** and **overflow: auto**             |  **position: absolute** and **overflow: hidden** | **position: fixed** and **overflow: auto** - without transform works as expected
:-------------------------:|:-------------------------:|:-------------------------:
Screenshot 2021-06-04 at 11 54 43  |  Screenshot 2021-06-04 at 11 55 10 | Screenshot 2021-06-04 at 11 54 09

### Details
It doesn't matter if `transform` triggers separation of parent into a separate rendering layer or not. Even `transform: translate(0)` that doesn't trigger separate layer disables `position: fixed`. At the same time `will-change: opacity` which does trigger layer separation still works with `position: fixed`


src/components/helpers/helpers.ts
```diff
@@ -144,3 +144,25 @@ export const safeSetAttribute = (element: Element, attr: string, value: string):
   if (element.getAttribute(attr) === value) return;
   element.setAttribute(attr, value);
 };
+
+/**
+ * Retrieves all parent elements including slots and pierced shadow doms:
+ */
+export const elementParents = (element: Element): Element[] => {
```
homer0, line 151:
```suggestion
export const getElementParents = (element: Element): Element[] => {
```

src/components/zen-popover/zen-popover.tsx
```diff
@@ -241,15 +241,28 @@ export class ZenPopover {
     return [offset, preventOverflow];
   }
 
+  parentsWithTransforms(popperTarget: Element): Element[] {
```
homer0, line 244:
```suggestion
  getPrentsWithTransforms(popperTarget: Element): Element[] {
```

src/components/zen-popover/zen-popover.tsx
```diff
@@ -241,15 +241,28 @@ export class ZenPopover {
...
+    const parents = elementParents(popperTarget);
+    return parents.filter(el => window.getComputedStyle(el).transform !== 'none');
```
homer0, line 246:
```suggestion
    return getElementParents(popperTarget).filter(el => window.getComputedStyle(el).transform !== 'none');
```

src/components/zen-popover/zen-popover.tsx
```diff
@@ -1,6 +1,6 @@
 import { Component, Host, h, Element, Prop, Watch, State, Event, EventEmitter, Method } from '@stencil/core';
 import { createPopper, Placement, Offsets } from '@popperjs/core';
-import { getComposedPath, waitNextFrame, applyPrefix } from '../helpers/helpers';
+import { getComposedPath, waitNextFrame, applyPrefix, elementParents } from '../helpers/helpers';
```
homer0, line 3:
```suggestion
import { getComposedPath, waitNextFrame, applyPrefix, getElementParents } from '../helpers/helpers';
```

src/components/zen-popover/zen-popover.tsx
```diff
@@ -241,15 +241,28 @@ export class ZenPopover {
...
+
   async createPopper(): Promise<void> {
     this.zenBeforeVisibleChange.emit();
 
     const popupWrap = this.host.shadowRoot.querySelector('.popup-wrap') as HTMLElement;
 
+    const errorParents = this.parentsWithTransforms(popupWrap);
+    if (errorParents.length) {
+      console.warn(
+        'Popover error! No popover parent should have css prop `transform` set! The following parents have css transform set:',
```
homer0, line 257: - It's either an error (because of the message) or a warning (because of the method) :P.
- Use `zen-popover` or prefix the whole thing with `zen-ui:`.

homer0 (commented):
👏 👏 ... amazing debugging 🤘!!!

- Only two tiny comments (one is in multiple places, that's why so many suggestions).
- Can you try enabling the vrt we disabled before?

chavlji:
@homer0 I've tride to rerun vr tests for dropdown, but the issue still persists :(

Screenshot 2021-06-04 at 15 29 14

Seems like something else is causing displacement there. We'll try to solve it in a different ticket.

chavlji:
Btw, when this is merged, we need to update zen-ui package in both execution and details app.

homer0:
@chavlji damn; did you tried the branch with the `mf-wizard` app?

homer0 (approved):

chavlji:
@homer0 I've tried it inside the app yeah.

PR #371, opened 2021-06-15
PLAT-2547: [mf-execution] Research performance issues on table

chavlji:
Adding new prop to popover that will enable us to wait with popover render until popover is rendered.

[JIRA ticket](https://reciprocitylabs.atlassian.net/browse/PLAT-2527)

### DIFF
--- a/src/components/zen-dropdown/zen-dropdown.tsx
+++ b/src/components/zen-dropdown/zen-dropdown.tsx
@@ -268,8 +268,9 @@ export class ZenDropdown {
     this.opened = this.popover.visible;
   }
 
-  onBeforeToggle(): void {
+  onBeforeToggle(): Promise<void> {
     this.matchDropdownWidth();
+    return new Promise(resolve => resolve());
   }
 
   componentDidLoad(): void {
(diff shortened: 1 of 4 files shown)


src/components/zen-dropdown/zen-dropdown.tsx
```diff
@@ -268,8 +268,9 @@ export class ZenDropdown {
     this.opened = this.popover.visible;
   }
 
-  onBeforeToggle(): void {
+  onBeforeToggle(): Promise<void> {
     this.matchDropdownWidth();
+    return new Promise(resolve => resolve());
   }
```
homer0, line 274:
```suggestion
  async onBeforeToggle(): Promise<void> {
    this.matchDropdownWidth();
  }
```

1. For an already resolved promise, you could use `return Promise.resolve();`
2. But if you add `async`, the method/function gets automatically wrapped in a promise, and all promises return a new promise, so there's no need to manually return one.

homer0 (commented):
Sorry, but I don't understand how a new callback prop would be the solution, why the listener was removed, and how the implementation would look like.

We can probably discuss this during tomorrow offline.

chavlji:
@homer0 I'll explain it on a daily.

homer0 (approved):
I would say, fix the tests and go for it :P.

PR #373, opened 2021-06-18
QUICK: zen-notification - Set inner html instead of text

chavlji:
Marko needs to insert link into notification:
image


mihaelpalfi (approved):
LGTM

